PR #9, opened 2015-06-26
Improved Trilateration UI

SamMackrill:
Quality of life, added splitter, added anchors and made editing cells less clunky


finwen:
Yes something is strange. Will investigate.

PR #30, opened 2015-11-07
Added functionality to allow a colour to be added to visited systems...

iainross:
... via a context menu on the Travel History grid which is then used for the 3D Star Map.  A new option in the Tools menu allows the colour for new systems to be specified.

Only known issue is that in the first session where the DB upgrade happens the code to group visited systems by map colour resolves to an Object Not Set error which results in visited systems not being highlighted in any colour.  It never happens again after restarting and I can't figure out what's causing it (if you can see what it is then I'd love to know...).

Let me know if there's anything else you want adding before merge or anything you want doing differently.


finwen:
Looks interesting will test this soon.

PR #37, opened 2015-12-29
Allow devs to optionally skip over slow EDSM updates when in Develop build

gregmalcolm:
So, I was about to start hacking when I realized waiting on the full EDD Load process takes rather a long time. So... I've put together a proposed where in Debug build only a developer can flip a setting which allows them to skip over those pesky slow EDSM updates.

In my hubris I also made a number of minor tweaks while I was in the codes. I can go back and remove any you don't like:
* Fixed a bug with "Log EDSM requests" checkbox not getting populated correctly on startup
* Renamed `Form1_Load` event to `EDDiscoveryForm_Load`. I tweaked the setting to load through program.cs to enable this to work.
* Renamed EDDConfig's `db` instance variable to `_db` to be match the code style of other instance variables.
* Added a `LogLine` delegatation variant in EDD for when theres a single argument.

### DIFF
--- a/EDDiscovery/EDDiscovery.csproj
+++ b/EDDiscovery/EDDiscovery.csproj
@@ -51,6 +51,7 @@
     <DefineConstants>TRACE</DefineConstants>
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
+    <UseVSHostingProcess>false</UseVSHostingProcess>
   </PropertyGroup>
   <PropertyGroup>
     <ManifestCertificateThumbprint>6586052CF2C6A2D8348CC25AA17842FA028AED2F</ManifestCertificateThumbprint>
@@ -66,6 +67,13 @@
   </PropertyGroup>
   <PropertyGroup>
     <ApplicationIcon>fge-logo-001-dark.ico</ApplicationIcon>
+  </PropertyGroup>
+  <PropertyGroup>
+    <StartupObject>
+    </StartupObject>
+  </PropertyGroup>
+  <PropertyGroup>
+    <NoWin32Manifest>true</NoWin32Manifest>
   </PropertyGroup>
   <ItemGroup>
     <Reference Include="Newtonsoft.Json, Version=6.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed, processorArchitecture=MSIL">
--- a/EDDiscovery/EDDiscoveryForm.cs
+++ b/EDDiscovery/EDDiscoveryForm.cs
@@ -89,7 +89,7 @@ namespace EDDiscovery
             tabControl1.SelectedIndex = 0;
         }
 
-        private void Form1_Load(object sender, EventArgs e)
+        private void EDDiscoveryForm_Load(object sender, EventArgs e)
         {
             try
             {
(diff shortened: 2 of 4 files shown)


EDDiscovery/EDDiscovery.csproj
```diff
@@ -51,6 +51,7 @@
     <DefineConstants>TRACE</DefineConstants>
```
gregmalcolm: All this wierdness is showing up because I tweaked the `Startup Object` property to go through `program.cs`. Because I renamed `Form1_Load` to `EDDiscoveryForm_Load`.

EDDiscovery/EDDiscoveryForm.cs
```diff
@@ -139,6 +139,12 @@ private void Form1_Load(object sender, EventArgs e)
 
                 textBoxEDSMApiKey.Text = db.GetSettingString("EDSMApiKey", "");
                 checkBox_Distances.Checked = eddConfig.UseDistances;
+                checkBoxEDSMLog.Checked = eddConfig.EDSMLog;
+
+                checkboxSkipSlowUpdates.Checked = eddConfig.CanSkipSlowUpdates;
+#if DEBUG
+                checkboxSkipSlowUpdates.Visible = true;
+#endif
```
gregmalcolm: I can't think of a good reason for normal users to skip the EDSM loading process, so I've hidden the checkbox from them.

EDDiscovery/EDDiscoveryForm.cs
```diff
@@ -235,10 +240,11 @@ private void EDDiscoveryForm_Shown(object sender, EventArgs e)
                 // Check for a new installer    
                 CheckForNewInstaller();
 
+                LogLine($"{Environment.NewLine}{Environment.NewLine}Loading completed!{Environment.NewLine}");
```
gregmalcolm: I thought it might be nice to let the user now that we're done. Yeah, I know the indicator goes away, but sometimes I start hitting EDSM update and am left uncertain when one process stopped and the other started.

Ok, now I think about that update is probably threaded but probably still a nice extra indication.

EDDiscovery/EDDiscoveryForm.cs
```diff
@@ -294,6 +300,15 @@ private bool DownloadMapFile(string file)
                 return false;
         }
 
+        private bool CanSkipSlowUpdates()
+        {
+#if DEBUG
+            return eddConfig.CanSkipSlowUpdates;
+#else
+            return false;
```
gregmalcolm: Running in Release build? No skipping for you!

EDDiscovery/EDDiscoveryForm.cs
```diff
@@ -399,12 +414,19 @@ private void GetEDSMSystems()
...
+                    LogLine("Skipping loading updates (DEBUG option).");
+                    LogLine("  Need to turn this back on again? Look in the Settings tab.");
+                }
+                else
+                {
+                    string retstr = edsm.GetNewSystems(db);
+                    Invoke((MethodInvoker)delegate
+                    {
+                        TravelHistoryControl.LogText(retstr);
+                    });
```
gregmalcolm: The "diffs" are a little hard to follow here, but its the `edsm.GetNewSystems(db)` part that getting optionally skipped over.

EDDiscovery/EDDiscoveryForm.cs
```diff
@@ -679,6 +700,19 @@ public void LogText(string text, Color col)
             }
         }
 
+        public void LogLine(string text)
```
gregmalcolm: Added the one arg variant

EDDiscovery/EDDiscoveryForm.cs
```diff
@@ -679,6 +700,19 @@ public void LogText(string text, Color col)
...
+            try
+            {
+                Invoke((MethodInvoker)delegate
+                {
+                    TravelHistoryControl.LogText(text + Environment.NewLine);
+                });
+            }
+            catch
+            {
+            }
```
gregmalcolm: Suggestion: I think we would be better off not eating the exceptions on these log methods. The chance of an unexpected failure is low here I think. But if something bad does happen, it would probably better to learn of it sooner rather eventually realizing that the logging has stopped. I've left it intact for now though.

PR #40, opened 2015-12-30
3d map singularity event

gregmalcolm:
**Main change**
Mostly this PR was about making the 3D map object a singleton item (wrapped by a proxy class with the exciting name: `3DMap.MapManager`). This means:
* Clicking on Show Map multiple times just causes the same form to refresh and reappear. So you get 1 map to rule them all.
* Closing the form just causes it hide

I've taken into account that there is a main map button and the trilateration flavor map button. Switching back and forth doesn't these 2 flavors does not seem to break anything as far as I can tell.

The `MapForm` constructors are a little different now: I took away all the arguments in `FormMap`. It's appearance is governed by setting properties and calling `Prepare()`. This will allow us to spawn the map in different ways for different contexts.

Mostly this is a foundation for upcoming changes, but one new feature: If a system with an EDSM point is selected in the Grid, the map will center on that item when the map appears.

Note: At the moment if you click on a system with no EDSM point and click on map it'll just default to the "default" center star. I'd like to make another adjustment soon which will make it display the system name in the text box (maybe in blue), but show coordinates with question marks so that it more or less mirrors the front page coordinates panel and gives the user a clue as to why it couldn't zoom in on their system.  

**Refactoring**

I'm afraid I got a little carried away with the refactoring from time to time. Which makes the diffs a little hard to read through. Mostly I:
* Broke up EDDiscovery_Form into smaller pieces. So there's less to wade through and the function names kind of act like live comments.
* Renamed instance variables/properties to match MS code standard. I figured there were a variety of styles in play so I just picked one.
* I converted a bunch of public vars to properties. The distinction is subtle but it just feels more kosher somehow.

Let me know if I'm going to far with these refactoring changes! I know I should be more restrained when I'm in someone elses codease but I seem to have poor restraint... :D

### DIFF
--- a/EDDiscovery/3DMap/FormMap.cs
+++ b/EDDiscovery/3DMap/FormMap.cs
@@ -40,22 +40,21 @@ namespace EDDiscovery2
         private const int ZoomMax = 100;
         private const double ZoomMin = 0.005;
         private const double ZoomFact = 1.2589254117941672104239541063958;
-        private readonly AutoCompleteStringCollection _systemNames;
-        bool loaded = false;
+        private AutoCompleteStringCollection _systemNames;
+        private SystemClass _centerSystem;
+        private bool loaded = false;
 
-        float x = 0;
-        float y = 0;
-        float z = 0;
-        float zoom = 1;
-        float xang = 0;
-        float yang = 90;
+        private float x = 0;
+        private float y = 0;
+        private float z = 0;
+        private float zoom = 1;
+        private float xang = 0;
+        private float yang = 90;
 
         private Point MouseStartRotate;
         private Point MouseStartTranslate;
-        private SystemClass CenterSystem;
 
         public List<SystemClass> StarList;
-        public List<SystemPosition> visitedSystems;
         public List<SystemClass> ReferenceSystems;
         private Dictionary<string, SystemClass> VisitedStars;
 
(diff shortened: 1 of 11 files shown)


EDDiscovery/3DMap/FormMap.cs
```diff
@@ -63,29 +62,79 @@ protected override void OnRenderButtonBackground(ToolStripItemRenderEventArgs e)
...
+                    _centerSystem = value;
+                }
+                else
+                {
+                    _centerSystem = null;
+                }
+            }
+        }
+
+        public String CenterSystemName
```
gregmalcolm: Useful for opening the map on a specific system with only a system name to go on.

EDDiscovery/3DMap/FormMap.cs
```diff
@@ -63,29 +62,79 @@ protected override void OnRenderButtonBackground(ToolStripItemRenderEventArgs e)
...
+            InitializeComponent();
         }
 
+        public void Prepare()
+        {
+            if (CenterSystem == null)
+            {
+                var db = new SQLiteDBClass();
+                string defaultCenter = db.GetSettingString("DefaultMapCenter", null);
+                OrientateMapAroundSystem(defaultCenter);
```
gregmalcolm: "Orientate" updates the controls before centering

EDDiscovery/Trilateration/TrilaterationControl.cs
```diff
@@ -731,12 +731,12 @@ private void toolStripButtonMap_Click(object sender, EventArgs e)
...
-                ReferenceSystems = CurrentReferenceSystems.ToList(),
-                visitedSystems = _discoveryForm.visitedSystems
-            };
-            map2.Show();
+            var map = _discoveryForm.Map;
+
+            map.Instance.Reset();
+            map.Instance.CenterSystem = centerSystem;
+            map.Instance.ReferenceSystems = CurrentReferenceSystems.ToList();
+            map.Show();
```
gregmalcolm: See how the map is interacted with now? First we set a few properties in the FormMap direct, then call the "show" method in the proxy class (`MapManager`)to manage the process of recalling the map.

EDDiscovery/3DMap/MapManager.cs
```diff
@@ -0,0 +1,35 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Windows.Forms;
+using EDDiscovery;
+
+
+namespace EDDiscovery2._3DMap
+{
+    public class MapManager
```
gregmalcolm: This is the map proxy I spoke of. It's mostly just there to manage the `show` through the MapForm instance.

Red13x3:
Mate, you are doing a great job that some may say is a bit anal, but it's what has to be done every once in a while to get everyone's code on the same page and in line with best practices. 

I'll be the first to say "Thank you!" For your enthusiasm and efforts!! :)

Mark

--
Mark Jelic
iPhone: 0402 322 901

PR #41, opened 2015-12-30
Better 16pix variant for application icon

gregmalcolm:
I was noticing that the 16x16 icon in the top left of the main form was a bit broken up. For example the 001 part was hard to read. So I copied the middle of the image from the 24x24 version into the 16x16 size, only sacrificing outer whitespace in the process.


gregmalcolm:
Went ahead and merged it. Just using this PR for the easier merge.

PR #44, opened 2016-01-03
Implement full camera transformations

gregmalcolm:
Got all the Camera transforms worked out! :D
So the only dataset's now married to to the CenterSystem are CenterSystem point and the Trilateration lines.

New features/side effects:
* The Camera's position and zoom level is now visible in the status bar
* Implemented a whole bunch of keybindings for movement and rotation. Will restrict the less useful rotation key bindings to DEBUG build later, because the user doesn't really need to know about them.
The main keys are WASD/Arrow keys to translate. +/- to zoom.
* Movement speed is the same whether you're zoomed in or out now.
* Map now defaults to Maximized. I feel it's the more useful default so that the glControl is immediately under the mouse cursor where it can receive keyboard commands.
* There was a small bug with the Z axis being inconsistant with offsetting the CenterSystem. No longer applies.
* Z Axis for points is no longer inverted.

Still got lots of improvements to roll in, but this is the first time things have been stable in a a couple of days...

### DIFF
--- a/EDDiscovery/3DMap/FormMap.cs
+++ b/EDDiscovery/3DMap/FormMap.cs
@@ -2,7 +2,6 @@
 using EDDiscovery.DB;
 using EDDiscovery2.DB;
 using EDDiscovery2._3DMap;
-using EDDiscovery2.Trilateration;
 using OpenTK;
 using OpenTK.Graphics;
 using OpenTK.Graphics.OpenGL;
@@ -45,8 +44,6 @@ namespace EDDiscovery2
         private bool _loaded = false;
 
         private float _zoom = 1.0f;
-        private float _xAng = 0.0f;
-        private float _yAng = 0.0f;
 
         private Vector3 _cameraPos = Vector3.Zero;
         private Vector3 _cameraDir = Vector3.Zero;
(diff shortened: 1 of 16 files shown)


EDDiscovery/3DMap/FormMap.cs
```diff
@@ -501,62 +667,63 @@ private void toolStripButtonGrud_Click(object sender, EventArgs e)
...
-        /// <param name="sender"></param>
-        /// <param name="e"></param>
-        private void glControl1_Paint(object sender, PaintEventArgs e)
-        {
-            if (!_loaded) // Play nice
-                return;
+            _loaded = true;
 
-            GL.Clear(ClearBufferMask.ColorBufferBit | ClearBufferMask.DepthBufferBit);
+            Application.Idle += Application_Idle;
```
gregmalcolm: Added this event so we could make more of a "GameLoop" rather than just having the GLControl repaint whenever it feels the canvas has changed.

Red13x3:
Well done, mate! This project has really leapt ahead since you came on board. Thanks, from a user that uses it EVERY time I play and right now, I'm at Sag A* and ready to head back.

If I could put in a request that the focus be put on storing the data (like the materials and making it easier to save the data) than displaying the saved data. Let's get some good data in there asap. :)

Mark

--
Mark Jelic
iPhone: 0402 322 901

gregmalcolm:
Heh, it's mostly the xmas break to blame for the progress.

I can't really pitch in with the materials stuff at the moment. Robert is trailblazing that one and I'd just get in his way right now. The screenshots I've seen are pretty cool though... :D

Issue #55, opened 2016-01-06
3D map is mirrored to the galaxymap ingame


muthan:
the 3D map has not the same presentation of the galaxy as the galaxymap ingame. It is mirrored over the horizontal layer.
So in my understanding the 3D map shows the inverse of the y values of the star coordinates

finwen:
Yes it looks like its mirrored upside down now.

gregmalcolm:
Really? That's very odd!

This is what I see 2.6.4.0 when I open the map:

![map](http://i.imgur.com/CIqaipH.png)

Which I believe matches the game. So are you both seeing this upside down?

I suppose it's possible there could be a camera bug. Which Star's are you starting from?

finwen:
That view is correct.
and it shows only X and Z values.
If you look at it from the side you see y values and then its upside down.



2016-01-06 16:38 GMT+01:00 Greg Malcolm <notifications@github.com>:

gregmalcolm:
Ah! I understand now, thanks! Yeah, side of effect of me making the points true coordinate representations of the real world coordinates. Guess we just need another world transformation that does the mirroring at the beginning of the rendering process.

muthan:
still an issue in 2.6.8

finwen:
Yes its inverted again.... strange

gregmalcolm:
Odd. I thought I'd fixed by adding an inversion of Z in the Scaling transformation in the rendering process:

```
           GL.Scale(_zoom, _zoom, -_zoom);
```

I can look into it when I'm next at my gaming PC, but probably won't happen till tomorrow.

gregmalcolm:
Ok, finally sorted it out

PR #58, opened 2016-01-06
Inverted the Z axis

gregmalcolm:
Fixed the Z axis by inverting the zoom (#55).

I confirmed Crab Pulsar and checking it was below Sol.
Zoom keys still seem to work ok, but wasn't able to confirm mouse wheel is still working:

Note: When testing the in map Center feature isn't working right now. I hope to fix that soon. In the meantime, you an test by filtering on the list of systems and clicking the 3d Map when filtered on the desired test point.


finwen:
Fixes #55

Issue #59, opened 2016-01-06
Center button in map is broke


gregmalcolm:
The "Center" button on the map is broke from my camera changes still. Need to adjust it to to work with the new dataset mappings.

Shouldn't be hard to fix, just haven't gotten to it yet.

gregmalcolm:
Sweet, I was getting really distracted with ed-materializer, so glad someone jumped on that one!

gregmalcolm:
Just got a chance to try this out. Much better, nice work!

PR #66, opened 2016-01-13
3D Map Settings

iainross:
Added 3D map controls to settings tab for Home System, Default Zoom and open on selection/home system toggle.

Travel History selection on opening the map now checks if the selection has coordinates and uses the last one prior to it that does if it doesn't.

3D map opens with the zoom and centre as defined above, Set Default button removed, buttons to re-centre on the Home System and Travel History selection added.


iainross:
Should also fix https://github.com/EDDiscovery/EDDiscovery/issues/59

Issue #67, opened 2016-01-13
Pressing Center home -> Null Reference exception


finwen:
I dont have a home system set.   

            ISystem sys = SystemData.GetSystem(_homeSystem);

sys will be null.   and i get a null reference exception then in  
            OrientateMapAroundSystem(sys);

iainross:
It's *supposed* to use Sol as a default if you've not set one.  Will check why it's not this evening and sort it.

iainross:
Aha, is uses Sol as a default on the settings tab but not in the 3d map cos I typed null instead and didn't notice - if you close it then it will save the Sol default and you'll have one set after that.

gregmalcolm:
Yep. You probably just need need to restore the original else so it can deal with looking for a player's own Sol or creating a temporary Sol if they haven't visited. You probably want a more if/else/else logic such as:

```
if We have a starting point use it
else if we have a _homeSysetm use it
else Use the Sol point or Make a temporary sol if we don't have one
```

iainross:
Now uses Sol (or dummy system at 0,0,0 if that's not there) if home system is null or invalid.

PR #68, opened 2016-01-13
Home button no longer errors with null or invalid Home System

iainross:


iainross:
Resolves https://github.com/EDDiscovery/EDDiscovery/issues/67

PR #70, opened 2016-01-17
Production ed-materializer server is now AWS based

gregmalcolm:
Configured ed-materializer server string to use the Amazon based hosting for the Release build. The old Heroku server will now be a sandbox for testing in the Debug build.

I've not seeded any dummy data into the Release database, so it will initially have zero rows.


finwen:
Maybee we should chang so we chan choose between production and test db then we are running in debug mode?

PR #71, opened 2016-01-17
Added save button to 2D Map viewer

iainross:
Saves jpg/bmp/png of 2D maps with travel history drawn on


gregmalcolm:
Sounds useful! Though we could probably simplify the interface a little by choosing the format ourselves. PNG is compact, web friendly and lossless. The user could always convert it later in MSPaint if they really wanted to. 

Not a blocker for this PR but just a thought. :)

finwen:
A copy image to clipboard could be good too

Issue #74, opened 2016-01-21
BITMAP convert - exception raised 


robbyxp1:
If you turn the delete original after conversion, it raises an exception.

This is because the BITMAP object keeps the file locked until its disposed.

quote from documentation BITMAP: Open from file:
The file name and path can be relative to the application or an absolute path. Use this constructor to open images with the following file formats: BMP, GIF, EXIF, JPG, PNG and TIFF. For more information about supported formats, see Types of Bitmaps. The file remains locked until the Bitmap is disposed. 

And there is another problem, the watcher is starting BEFORE the systems are loaded. This means if you hit take screenshot too quickly (as I did during debugging), you will get an exception on imagehandler.cs:131 ISystem cursys = _discoveryForm.TravelControl.GetCurrentSystem();

To fix:
imagehandler.cs:54 remove startwatcher at the end of InitControl
eddiscoveryform.cs:147 add imageHandler1.StartWatcher(); (just below Text +=)

Finally to fix the remove problem, in imagehandler.cs:

.. code
                    else
                    {
                        ED_PIC.Save(output_folder + "\\" + new_name + pic_ext, System.Drawing.Imaging.ImageFormat.Png);
                    }

                    ED_PIC.Dispose();   // need to release lock on files..
                    Screenshot_PIC.Dispose();

                    if (checkBoxRemove.Checked) // Remove original picture
                    {
                        File.Delete(e.FullPath);
                    }

                    Invoke((MethodInvoker)delegate { TravelHistoryControl.LogText("Converted " + e.Name + " to " + new_name + pic_ext + Environment.NewLine ); });
}
The last Invoke provides a handy confirmation that the image was converted... Not strictly necessary but useful.

Rob

gregmalcolm:
Thanks for researching this! Any interest in putting in a Pull Request for this? It'll get your name on the contributors list... ;)

robbyxp1:
i work, but the repository I'm working on has a bigger change to the route planner tab. Can I do a pull request just of one file?

robbyxp1:
I would I mean ;-)

robbyxp1:
Ignore the pull requests I created.. Its not working as I think it does.

gregmalcolm:
Absoluetly! Tou can create pull requests with commits in them. The trick is to to push up different git branches. 

This sequence would  get you there:

Note: Commands given are to be run from git bash, assuming you are in the project folder. I'm also assuming you are working off your own github fork:

[...]

PR #77, opened 2016-01-21
Fiximagehandler

robbyxp1:
Fix the image handler..

Fix the problem with deleting the file.

Hopefully this is right


robbyxp1:
Try this, this looks good..

gregmalcolm:
Woot!

robbyxp1:
;-) Bit confusing for an old git like me.. I'll read up more about how GIT works.

gregmalcolm:
Fixes #74

gregmalcolm:
Just checked over the code. I'll let someone more familiar with this aspect of the code decide whether or not to merge it, but looks fine to me.

robbyxp1:
i did a lot of testing on it today, and the problem disappeared. As I said in the previous post, the documentation states that the file is locked until the resource is deallocated. So we need an explicit call to force it to unlock the file.  The last bit is a nicety and it provides a visual confirmation the screenshot was converted.

finwen:
Looks good.  Will merge and test.

Issue #86, opened 2016-01-27
Feature Request: Fast-travel mode - focus "Distance to Next System" when Windows becomes active


mwerle:
During Fast-travel where a long route is entered I only enter the distance to the next system (and trilaterate every 10th system or so).

So, jump into system, honk, turn to next system, start up FSD, throttle-full, check galaxy-map for distance to next system (2 decimals), alt-tab to EDDiscovery, use mouse to select "Distance to next System", enter data, alt-tab back to ED.

It would be nice to be able to switch the App into a "Fast Travel" mode, in which the App focuses on the "Distance to Next System" input field when the window becomes activate. That way I wouldn't have to use the mouse during jumping.

Even better: have a hotkey which pops up a small "Distance to Next System" dialog. No need to alt-tab etc.

Red13x3:
Yup, I suggested this exact same thing a long while back. Like I think it's request #15 or so. Not had any feedback or progress on it. 

But I'll add a +1 to this, as it supports mine. ;)

--
Red_XIII

PR #89, opened 2016-01-27
Fast-travel mode (Distance to Next System)

mwerle:
Proof-of-concept to fix issue #86 

Add a new menu option "Fast Travel Mode".

When checked, the "Distance to Next System" textbox is automatically selected when the form is activated.

Works For Me.

### DIFF
--- a/EDDiscovery/EDDiscoveryForm.Designer.cs
+++ b/EDDiscovery/EDDiscoveryForm.Designer.cs
@@ -73,6 +73,8 @@
             this.adminToolStripMenuItem = new System.Windows.Forms.ToolStripMenuItem();
             this.forceEDDBUpdateToolStripMenuItem = new System.Windows.Forms.ToolStripMenuItem();
             this.syncEDSMSystemsToolStripMenuItem = new System.Windows.Forms.ToolStripMenuItem();
+            this.optionsToolStripMenuItem = new System.Windows.Forms.ToolStripMenuItem();
+            this.keepOnTopToolStripMenuItem = new System.Windows.Forms.ToolStripMenuItem();
             this.helpToolStripMenuItem = new System.Windows.Forms.ToolStripMenuItem();
             this.gitHubToolStripMenuItem = new System.Windows.Forms.ToolStripMenuItem();
             this.eDDiscoveryHomepageToolStripMenuItem = new System.Windows.Forms.ToolStripMenuItem();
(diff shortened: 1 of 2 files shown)


Red13x3:
That means EVERY time you go to the form, the focus will be there which yeah, works, but may piss people off if they are editing the Notes field and then flick back and forth. 

What I suggested in my post (all that while back) is that the focus goes to that field when EDD picks up that you have arrived in a new system. It does that even if it doesn't have the focus as I see it update the history list when I'm in witch space.

What I would like to happen is that when I then switch to EDD, the new system is selected and the cursor is already blinking in that "distance" field, ready for input.

Or better still, have a new field called "distance of next jump" that I fill in before I jump. When EDD registers that I've arrived in a new system, it takes that "next jump" data I put in and saves it as the distance from the previous system. This would stop the goldfish memory of mine that always seems to forget what numbers I need to enter and then having to go back to the Galaxy map, search the previous system and then type in the 2 digit number. :/

--
Red_13
iPhone: 0402 322 901

mwerle:
Not every time; only if you have the option enabled. People who get pissed off at this feature don't need to enable it. And yes, there's room for improvement; so far this is just a proof-of-concept I knocked up quickly.

Also it focuses on the "Distance to Next System" field, which is indeed filled in -before- the jump. I'm guessing you're talking about the "Distance From" field? I only ever use that if I got the timing wrong or forgot to fill in the "Distance To" field before the jump.

Having to fill in the last distance jumped takes too much time as you have to go to Galaxy map and search for the previous system in order to get the distance. Or you have to remember it from before the jump. You also have to press the "Update" button.

My fast-jumping workflow is as follows:
- Select next system in jump-chain if not already selected (joystick button)
- Start frameshift drive charging (joystick button) and align to vector
- Open Galaxy map (I currently have '5' mapped to it; I may use a dedicated joystick button at some point) to get distance to next system (Galaxy map is already focused on destination system so distance can be read instantly)
- Alt-tab to EDDiscovery (Elite is still visible, with the distance, so no need to remember it)
-  - previous to this commit, I had to use the mouse and click on Distance-to-Next-System
-  - with this commit (and this new option enabled) DtNS field is already selected
- Type in distance
- Alt-tab back to Elite in time for the jump sequence

When the jump completes, the distance has been automatically entered into the Travel History, and you're ready to honk-and-jump to the next system.

If I want to enter Notes, or Trilaterate a system, this "Fast Jump" routine doesn't work, so I have all the time in the world to select the correct GUI elements anyway.

I guess what this is is "Fast Travel TO" and what you're wanting is "Fast Travel FROM"

mwerle:
Here are the fields I'm talking about:
http://s10.postimg.org/5hjwrlgsp/EDDiscovery_Distance_Fields.png

Issue #115, opened 2016-02-20
Image Conversion - Problem in dialog creation, and add more options for filenames


robbyxp1:
1) There is an issue with the dialog setup- its not showing the selected filename properly in the example box.  

This is due to Init function calling CreateFileName, it trying to get the dialog combo box index but this is gated off by if (Create), which its not at the time of Init.

2) Commanders in discord have asked during chat for more options for filenames (a reason not to use this tool).  Lets give 'em what they want.

3) The fix for filenames what I pushed should really be in CreateFileName, not where I placed it. My fault.

I have a branch with the changes on, and I will test them Sunday and push a fix/enhancement.

robbyxp1:
4) There are accesses in the crop system to dialog items, which is not allowed since they are in another thread.
5)There is further problems with deleting the file if you crop.. I'm investigating.

robbyxp1:
9) Does not save preview option
10) can not crop and delete
11) did not show cropped preview
11) Windows date time option crashed on mine due to 22/10/2022 format of dates.

mwerle:
Thanks for the various fixes; sorry for the issues I introduced (I had tested my changes, but only within a limited set of scenarios).

As for **2) Commanders in discord have asked during chat for more options for filenames (a reason not to use this tool). Lets give 'em what they want**, I have thought about this as well, and the best solution (long-term) is to do the same as most other software packages; that is, add a set of tokens (filename, date (pre-determined as well as split into components), current system, pilot, etc etc) and let people create their own filename patterns, with a set of default ones.

robbyxp1:
No problem, its hard to check all the scenarios, it took me ages ;-)

So, I agree, a configurable system would be better, but that would have meant more testing, so I just did the simple thing and put a fixed set of options in there.

mwerle:
Yes; for not it's great as-is. I was just noting down thoughts for the future. Having a tokenised system will require a LOT more work!

robbyxp1:
Well, not so much work, as you could just pass the string format given by the user to Date.ToString, after doing come work doing your own replacements.  

The normal date/time formats YYYY MM DD mm dd etc could be just passed straight thru, then we could add on some of our own (processed before being passed to tostring) like %S = System.

The biggest problem with it is 1) Testing the combos and 2) Explaining to the user how it works.  2) is probably the worse.

robbyxp1:
Closed - extra formats were added

PR #119, opened 2016-02-21
Imagefilenameoptions

robbyxp1:
(submitted prev pull - hopefully not a duplicate, but it never showed up)

Fixes issue #115.

Please extensively test, I've tested it with multiple combinations of fileformat, folders, cropping and preview.


robbyxp1:
Make sure we test it before deployment - its a big change. I took about an hour today trying different combos!

Issue #129, opened 2016-02-24
Enforce single-instance mode


mwerle:
Instead of opening a new instance of the app, ensure only a single copy can run at a time. (Sorry, can't think of the proper term for it right now)

Open currently-running copy, pop up a dialog, whatever. Maybe it's just me, but several times now when shutting down the computer I've found a second copy of EDDiscovery running.  I'm actually amazed my DB hasn't blown up!

finwen:
+1

But we need to be sure that the other EDD is alive  and not a half dead or
just a thread that have not shut down and UI is closed.
EDD had a few bugs before then a thread didnt stop at exit and the proces
thas still running in background.  In that case we cant  move the old to
front.

One other way to do it can also be to "kill" the old EDD.

2016-02-24 4:52 GMT+01:00 Michael Werle <notifications@github.com>:

PR #130, opened 2016-02-24
Mw feature singleton

mwerle:
I have no idea why it's complaining about whitespace AGAIN.

I merged in the upstream whitespace changes, now it's trying to revert them again?  I didn't see this in my client only now that I'm trying to do this pull request.

I've also tried doing a complete new checkout of the entire source tree. Really at my wits end why this keeps happening. If someone can provide me with their complete GIT configuration (sans keys, of course) maybe I can trace it as I appear to be the only one with this problem - on two separate computers.

Anyway, you want to just merge in this commit, which adds a helper class and wraps the app startup with a global mutex: https://github.com/EDDiscovery/EDDiscovery/commit/9ce7956a7fc14f12b0216f5abc1e68fff0dfc677

There's additional things to add (such as popping up a dialog or bringing the app window to the foreground, but this is an initial quick-fix for #129.


mwerle:
Actually; I can't let this rest. I'll kill this pull request and try to figure out this bleeping whitespace issue first.

I mean, my master branch _SHOULD_ be a perfect copy of the upstream one!  I do all my work in a separate branch (due ot VS changes), and create clean feature branches for individual tasks.

Where do the whitespace issues come from???

mwerle:
https://answers.atlassian.com/questions/328640/whitespaceline-endings-even-with-.gitattribute-file

Issue #134, opened 2016-03-06
Improvements to the Route finder AStar method


robbyxp1:
Improve the star selection algorithm further to pick stars withing a "barrel" of the vector between the two points.  

Initially pick a 10% margin barrel, min 10ly, max 250ly (so a 300ly trip would accept stars within 30 lys of the vector, 50 ly would have a 10ly barrel).

If that fails (and it might do, relax the 250 ly limit to 500, then 1000 and retry).

This way it might work with enough speed in deep space with 900ly jumps.

robbyxp1:
Coded today - its still too slow for use. Even with a minimum jump size set to 50% of max range, and doing a  much better job of picking the initial star list, the arc creation and search takes too long for a Sol->SagA* search.

We can 1) commit this (no real harm - makes it faster in civ space) 2) Abandon 3) delete the AStar altogether.

robbyxp1:
Sorry, should stay open

robbyxp1:
See #137 removed alogorithm and tidied.

Gerdofal:
I'm not sure if this will help you since I'm doing all my work in SQL called by PHP, but here is a query that returns all the starts in a cylinder defined by the start and end points and a radius. I get results, even on a massive radius, within seconds.

`
//These test coordinates are SOL to Beagle Point.

//Start system
$ax = 0;
$ay = 0;
$az = 0;

//End System
$bx = -1111.5625;
$by = -134.21875;
$bz = 65269.75;

//Calculation of difference
$cx = $bx-$ax;
$cy = $by-$ay;
$cz = $bz-$az;

//Radius of column
$radius = 250;

//Query
$query = "
	SELECT
			name,
			x,
			y,
			z,
			(((x*$cx+y*$cy+z*$cz)-($ax*$cx+$ay*$cy+$az*$cz))/($cx*$cx+$cy*$cy+$cz*$cz)) as tmin,
			($ax + ((((x*$cx+y*$cy+z*$cz)-($ax*$cx+$ay*$cy+$az*$cz))/($cx*$cx+$cy*$cy+$cz*$cz)) * $cx)) as lx,
			($ay + ((((x*$cx+y*$cy+z*$cz)-($ax*$cx+$ay*$cy+$az*$cz))/($cx*$cx+$cy*$cy+$cz*$cz)) * $cy)) as ly,
[...]

PR #137, opened 2016-03-16
Removed AStar Algo

robbyxp1:
Astar removed, more checks on selected index of metric to prevent
faults, removed unused code.


robbyxp1:
See issue #134

PR #141, opened 2016-03-22
Perform log file parsing using a regular expression instead of hard-c…

mwerle:
…oded substring extraction.

Also fix issue when the system entry is not at the start of a line, which can sometimes happen.

To test, recommed : https://regex101.com/

```
{01:02:03} System:0(Sagittarius A*) Body:1 Pos:(1,2,3) Supercruise
{01:02:03} System:0(Some System (with) [brackets] {and} `stuff` 'in' @it@ #which# +should+ be *hard*) Body:1 Pos:(1,2,3) Supercruise
```

### DIFF
--- a/EDDiscovery/SystemPosition.cs
+++ b/EDDiscovery/SystemPosition.cs
@@ -4,6 +4,7 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
+using System.Text.RegularExpressions;
 using System.Threading.Tasks;
 
 namespace EDDiscovery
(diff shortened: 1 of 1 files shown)


mwerle:
Ok, deleted the original branch (due to a rather irate commit by myself, which I thought would remain private), recreated it, and it seems to have confused github somewhat. 

To recap: I've tested with some difficult system names, which include brackets, stars, and various other things. Obviously not to exhaustion, but somewhat confident it should work.

@gregmalcolm I wasn't suggesting a message box; I was thinking of the notification area at the bottom of the EDDiscovery window, or an actual log file. We'd need to track line numbers though.

gregmalcolm:
Tip, if you ever want to change history on a pull request, simply roll back your commits one at a time:

```
git reset --mixed HEAD~1
```

Make the change recommit, then force push then adjusted history:

```
git push -f my-pr-branch
```

And the irate comment was never there... :D 


Oh, and sending the warning to the progress window sounds like a good idea.

gregmalcolm:
Btw, should probably add the disclaimer: Never force push to Master. It'll mess up everyone who's branched off from it... :D

mwerle:
Hmm, tried to do that, wouldn't let me push, not even with -f, hence just delete and re-submit.  And yeah, I know never to muck about with master :)  Which is why I tend to do all my dev in a side-branch and merge into master when I'm happy.

PR #143, opened 2016-03-23
Removed last row from code

robbyxp1:
This fixes a crash when rich note text changed occurs without anything
selected (can happen if travelhistory is empty) during a theme change.
And removes any possibility of storing when we don't have a selected row

### DIFF
--- a/EDDiscovery/TravelHistoryControl.cs
+++ b/EDDiscovery/TravelHistoryControl.cs
@@ -238,7 +238,6 @@ namespace EDDiscovery
             
             if (dataGridViewTravel.Rows.Count > 0)
             {
-                lastRowIndex = 0;
                 ShowSystemInformation((SystemPosition)(dataGridViewTravel.Rows[0].Cells[1].Tag));
             }
             System.Diagnostics.Trace.WriteLine("SW3: " + (sw1.ElapsedMilliseconds / 1000.0).ToString("0.000"));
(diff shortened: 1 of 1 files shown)


EDDiscovery/TravelHistoryControl.cs
```diff
@@ -665,10 +658,11 @@ private void buttonUpdate_Click(object sender, EventArgs e)
                 distance.Store();
                 SQLiteDBClass.AddDistanceToCache(distance);
 
-                dataGridViewTravel.Rows[lastRowIndex].Cells[2].Value = textBoxDistance.Text.Trim();
+                if (dataGridViewTravel.SelectedCells.Count > 0)          // if we have selected (we should!)
+                    dataGridViewTravel.Rows[dataGridViewTravel.SelectedCells[0].OwningRow.Index].Cells[2].Value = textBoxDistance.Text.Trim();
```
gregmalcolm: Trivial: Personally I'd extract the `dataGridViewTravel.SelectedCells[0].OwningRow.Index` bit into a variable or method for better readability

gregmalcolm: Refactoring suggestion: Lots of references to "cells[2]", "cells[3]" etc. Maybe using an enum instead of a number would aid readability and also make the code more change resistant

gregmalcolm:
Overall the change looks fine when I eyeballed it.

I threw in a couple of refactoring suggestions, but that's just icing

Issue #116, opened 2016-02-21
Map Proposal: True Perspective vs. Isometric Perspective


RedfoxCh:
Besides being an awesome feature, something in the 3D map's representeation always seemed a little off to me. If I am not mistaken, the current 3D map view is rendered in an isometric perspective, which is not what our (my?) brain would naturally expect from a realistic depiction of a 3D object.

Is it possible to add a 'True Perspective' option when viewing the 3D map?

For reference, I googled this image which shows the difference between the two perspectives nicely: [https://myweb.rollins.edu/jsiry/tp.gif](https://myweb.rollins.edu/jsiry/tp.gif)

finwen:
No idea but its probably just some OpenGL setting

2016-02-21 11:26 GMT+01:00 RedfoxCh <notifications@github.com>:

gregmalcolm:
It's already Perspective. You may get tge illusion that it's not because of certain rendering options.

For example, the camera lense is pretty wide, the points are just rendering as solid colors rather than lit and the culling currently does wierd things when you get close.

RedfoxCh:
Hey Greg, thanks for elaborating. Possibly I'm not using the correct terms. In the 3D map screenshot below I've highlighted two parallel rows of systems, both about 50kly long. In my basic understanding of a perspective view these two rows should gradually converge (visually, towards the vanishing point) the further away they get from the viewer's position. On the current 3D map they are rendered in parallel, all the way.

[http://i.imgur.com/2B53p4r.png](http://i.imgur.com/2B53p4r.png)

I hope this makes sense... and if this is just me, feel free to shelve the issue :-)

gregmalcolm:
Maybe I'm wrong. Well, whether its perspective or other rendering quirks, I
think we can agree the rendering can be improved on

klightspeed:
In 3.0.4, one can now switch between parallel (or orthogonal) projection and perspective projection.

RedfoxCh:
Late to the party, sry - just wated to say the perspective view is fantastic, thanks!!

PR #154, opened 2016-03-30
Add perspective rendering on 3D map

klightspeed:
This adds the option for perspective rendering as an alternative to the current orthogonal rendering.

This was proposed in #116

This also maps Right-drag to Y-axis translation, and Left+Right-drag to panning on the X and Z axes.

### DIFF
--- a/EDDiscovery/3DMap/FormMap.Designer.cs
+++ b/EDDiscovery/3DMap/FormMap.Designer.cs
@@ -72,6 +72,7 @@ namespace EDDiscovery2
             this.glControl.Paint += new System.Windows.Forms.PaintEventHandler(this.glControl_Paint);
             this.glControl.MouseDown += new System.Windows.Forms.MouseEventHandler(this.glControl_MouseDown);
             this.glControl.MouseMove += new System.Windows.Forms.MouseEventHandler(this.glControl_MouseMove);
+            this.glControl.MouseUp += new System.Windows.Forms.MouseEventHandler(this.glControl_MouseUp);
             this.glControl.MouseWheel += new System.Windows.Forms.MouseEventHandler(this.glControl_OnMouseWheel);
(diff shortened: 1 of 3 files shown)


gregmalcolm:
My experience with working on the rendering sequence has taught me that the matrix transformation can be tricksy. I'll ackknowledge there was a bug with mouse drag after rotating the map already that I never got around to fixing it, but...

Can you confirm that moving the map around with the WASD keys still works? Even if the map is rotated and zoomed?

klightspeed:
`W` moves the focused point "up" from the camera's perspective, which is North when unrotated.
`S` moves the focused point "down" from the camera's perspective, which is South when unrotated.
`A` moves the focused point to the camera's left, which is West when unrotated.
`D` moves the focused point to the camera's right, which is East when unrotated.

`R` and `F` seem to be swapped - `R` goes rearwards, and `F` goes forwards, while the names of the actions suggest they should do the opposite.

These work correctly with respect to the rotation and zoom of the camera.  Zooming in the camera slows down the movement, and zooming out the camera speeds it up.  Motion is along the left/right/up/down/forward/back axes according to the camera's perspective.

Pitch up and down raises and lowers the camera's lattitude above / below the point.
Roll left and right rotates the map along its Y axis.
Yaw left and right is dependent on the pitch and roll.

gregmalcolm:
As long as the movement keys work in combination is the main thing. When the transformation sequence is bad movement becomes wierd... :)

Another gotcha is keeping the Y axis pointing the right way. I guess if you check that the DW path is ascending at Sag A rather than descending you're in the clear

klightspeed:
I did indeed check that the DW path at Sagittarius A* ascended, which is why I added 18a41ee.

For the camera Y, a quick fix might be to negate Y on line 149 of `FormMap.cs`, and negate it again when displaying the coordinates.

PR #160, opened 2016-04-02
3D Map: Click to select and double-click focus

klightspeed:
This adds the ability to see what star is at a given position in the 3D map, and to center the map on that star.

A single-click on a star will select the star, but not set it as the center system.

Double-clicking on a star will select the star, set it as the center system, and center the map on it.


mwerle:
Seems to work well!  Only comment I'd have is that centering the map on the double-clicked star also resets the map orientation.  Would be better user experience to leave the orientation.

finwen:
Looks great. Only did limited testing tonight but will test more later
today

Seems to work well! Only comment I'd have is that centering the map on the
double-clicked star also resets the map orientation. Would be better user
experience to leave the orientation.

—
You are receiving this because you modified the open/close state.
Reply to this email directly or view it on GitHub
<https://github.com/EDDiscovery/EDDiscovery/pull/160#issuecomment-204897517>

Issue #164, opened 2016-04-03
Multiple Commanders - Really doesn't work


Red13x3:
Has anyone tested this functionality?  I finally purchased my second commander and wanted to track his progress from the beginning... I set up a second commander in EDD, and even got a EDSM API key and all. It actually seemed to work OK, with me being able to flick between the two commanders, and them having two different travel histories.  But after logging off Elite and EDD one night and coming back to it a few days later, the two commanders travels are merged into one travel history (the original commander's) and the second commander having nothing.

Will this have screwed my original travel data?  Can I send any files in for someone to analyze?  I can certainly show screenshots.

klightspeed:
The commander ID is never set before adding the visited systems to the database.

Red13x3:
![2016-04-04_214926](https://cloud.githubusercontent.com/assets/13119623/14247222/044e8e3c-fab0-11e5-8f74-35114ccf881d.png)

Red13x3:
As you can see from the above, the records 91 and onwards are from my Red_13 character that is in the Black doing exploring... And my Red_XIII character is in the human bubble doing missions, shown from record 90 and lower.  If I select the Red_XIII commander, nothing shows up in the history. The original commander (Red_13) has all the data. :(

finwen:
Multi commander is limited and need allot of improvements. But you can move
data to another commander.

My shoulder interfered with my work on it allot but  now i  am slowly back
on working on EDD again. I have 2 commanders and will work on some
improvements soon :)

2016-04-04 14:00 GMT+02:00 Mark <notifications@github.com>:

Red13x3:
Hi Finwen... You mentioned being able to move data to another commander... Can you tell us how to do that?  Also, if I'm moving from one computer to another, can I safely grab the data file and move it to the new computer?

Red13x3:
When going to Add the second commander, the first commander is listed as being number "0" and when you add the new one, it is defaulted as number "2"... Yet when you restart the app, the commander 0 is actually duplicated to "1" as well, so you have three commanders listed.

Also, another problem: It does not remember which commander you had selected last before quitting and each time you start the app, it defaults back to the first commander you had created.

klightspeed:
> You mentioned being able to move data to another commander... Can you tell us how to do that?

Select the entries to be moved from the travel history list, right-click them, and click the "Move to another commander" option.

> Also, if I'm moving from one computer to another, can I safely grab the data file and move it to the new computer?

[...]

PR #169, opened 2016-04-06
Set active commander when saving netlogs

klightspeed:
This should prevent a recurrence of #164 - the commander wasn't being stored in the database.

Some sort of fixer tool would need to be used to associate netlogs with the correct commander(s).

### DIFF
--- a/EDDiscovery/EDSM/EDSMSync.cs
+++ b/EDDiscovery/EDSM/EDSMSync.cs
@@ -168,6 +168,7 @@ namespace EDDiscovery2.EDSM
                             vs.Time = system.time;
                             vs.MapColour = _defmapcolour;
                             vs.EDSM_sync = true;
+                            vs.Commander = EDDiscoveryForm.EDDConfig.CurrentCommander.Nr;
 
 
                             vs.Add();  // Add to DB;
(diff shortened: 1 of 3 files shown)


finwen:
Thajks.  That should iprove it in a good way.

Red13x3:
So in my logs where both commander data is merged into one... Can I manually edit something to correct the data?

I'm pretty savvy around this all, but I'd need to be told exactly what file to edit and what is he best editor to use. 

Thanks!

--
Mark
iPhone: 0402 322 901

klightspeed:
At the moment, you'd need to use SQLiteBrowser or something similar to execute some SQL UPDATE commands.

If you can figure out what date/time ranges the second commander was active (if you synced to EDSM before closing EDDiscovery, it should be recorded on EDSM), you could open `%LOCALAPPDATA%\EDDiscovery\EDDiscovery.sqlite`, and execute the following command in SQLiteBrowser for each of the time ranges your second command was active:
```
UPDATE VisitedSystems SET commander=1 WHERE Time >= '2016-xx-xx xx:xx:xx' AND Time <= '2016-xx-xx xx:xx:xx';
```
where `2016-xx-xx xx:xx:xx` is the start/end times when your second commander was active.

PR #180, opened 2016-04-21
Update to edmaterializer api v3 (JSON API)

gregmalcolm:
Lots changes for Prospecting:

1) Changed from "ActiveModelSerializer" json to "JSON API" json. So payloads come back and forth from edMaterializer now look like this: http://jsonapi.org/
2) Changed the ResponseData so the response.body gets streamed through when an error occurs. Sometimes with 422 errors there might be a validation error like `{world: [can't be blank]}`.Useful for populating error boxes. Also useful so you know why your requests are failing.
3) Broke up edmaterializer.cs into "repository" classes for reading and writing from API. I did this because it was getting crowded with Worlds, Stars and WorldSurveys already, and I needed to add Surveys and Basecamps.
4) Implemented/updated respositories and data classes (EDWorld, EDStar, etc) for Worlds, Stars, WorldSurveys and Surveys. Basecamps isn't done yet, but it can wait. Some details such as Material attributes are still not done.
5) Established assocications for loading WorldSurvey and Survey data from a World. Look for `GetWorldSurvey()` and `GetSurveys` methods in `EDWorld`. We'll need some more of that to link to Basecamps. Eventually.
6) I have reason to believe the current EDDiscovery installer isn't packaging the credential file (AppSettingsSecets.config). I say that because I can see authentication is failing. I've added something to the startup process to let us know if the credentials can't be found
7) I've hardcoded the EDMaterializer username/password for the dev server so other developers are able to test changes.

That is all.


gregmalcolm:
One other thing to note. The dev server is currently updated to v3 api:

http://ed-materializer.herokuapp.com/api/v3/worlds

You can tell because of the word "data" at the front. The production db is still on v2 for now. I'll upgrade it after this PR has gone through. So for comparison:

http://api.edmaterializer.com/api/v2/worlds

Issue #199, opened 2016-04-29
Feature Request: Ability to pull private comments from EDSM back into EDD


CmdrZach:
that it all.

klightspeed:
At the moment, EDDiscovery doesn't link notes to any commander, while EDSM private comments are linked to a commander.

Should notes be linked to a particular commander?  If not, what should happen if more than one of the user's commanders have a private comment on the system, or EDSM has a private comment on a system where EDDiscovery already has a note?  Should the most recent private comment take precedence, regardless of source?

CmdrZach:
Maybe select on import which commander to import them to? Also, they will be stored on EDSM under a commander name, so default to that? Or only allow import to the correct commander name or just whatever local commander is selected when import comments is chosen?

All I know is I enter comments in EDD as me, it exports them to EDSM on my commander name, and I would like to get them back: After  a rebuild and losing my comments, and now a corrupt database and losing all my new comments......

klightspeed:
Resolved in PR 200

Issue #156, opened 2016-03-30
Notes / Comments Field added to EDSM Website does not Sync with Desktop Application


codevolunt:
Are the Notes (Desktop) and Comments fields (EDSM Website) fields equivalent?

When I add a Comment via the website, it does not sync with the Notes field in the Desktop application. Is this intentional? Could they be synched in a future version?

EDIT:: Notes added to the Desktop application DO populate the Comments field on EDSM website, but it is the reverse that does not seem to be working.

PR #200, opened 2016-04-30
Sync comments from EDSM

klightspeed:
Sync comments from EDSM under the active commander.  Note that this currently uses the most recent comment, regardless of which commander the comment was filed under.  Requested in #156 and #199


finwen:
Funny. I talked about this function yesterday with another commander. And i planned to wrtie it next week.  Thanks for reading my mind!

PR #208, opened 2016-05-03
Portability fixes

klightspeed:
* Use `Path.Combine()` and `Path.DirectorySeparatorChar` in place of concatenating using backslashes
* The filename of `EdMaterializer.cs` was all lower-case, which caused a build failure under Linux
* Mono under Linux does not support Windows ACLs on mutexes.  Don't die if unable to set the ACL on the mutex.
* Use `Mono.Data.Sqlite` in place of `System.Data.SQLite` where available
* Change everything using `SQLite*` classes to use the equivalent `IDb*` interfaces instead
* Use `this.WndProc(ref Message)` instead of `user32::SendMessage(this.Handle, ...)`
* Use `this.Capture=false` instead of `user32::ReleaseCapture()`

With these fixes, EDDiscovery runs under Linux, and should also run under OS X.

Known issues under Linux:
* Menu strips are pink and only paint the last item
* Dropdown lists are dark
* Scroll bars don't appear to be rendered
* Many of the text boxes are over-sized
* Attempts to move or resize the window do not work when it is borderless
* Clicking the Close button fails to close the application
* Mono Mutexes apparently are non-functional on operating systems other than Windows - would need to use e.g. a lockfile to emulate mutexes on these systems.


Red13x3:
Great work, Ben. 

Would be awesome to have this work on the Mac as well. :)

Mark

--
Mark
iPhone: 0402 322 901

klightspeed:
I am re-working the Mono.Data.Sqlite commit to be more robust, so it should prefer System.Data.SQLite, and use Mono.Data.Sqlite when that doesn't work.

PR #214, opened 2016-05-05
Upgrade to EDMaterializer API v4

gregmalcolm:
Upgraded the EdMaterializer interface from v3 to v4 which uses JSON API payloads for POSTs and PATCHs.
Also:
* Implemented a method for populated message boxes when a JSON API request fails. So if a Save fails the user is now told why.
* Finished implementing EDBasecamp Model/ Basecamp repository
* Finished supplying methods for accessing associated models from any given model. For example from a survey, basecamp or world survey you get request the world that the model belongs to. From a world you can access list all assoicated surveys. From a basecamp you can list all basecamp surveys.


gregmalcolm:
@klightspeed makes a great point. Withdrawing the PR until I address this.

PR #215, opened 2016-05-05
Upgrade to EDMatieralier API v4 (take 2)

gregmalcolm:
Upgraded the EdMaterializer interface from v3 to v4 which uses JSON API payloads for POSTs and PATCHs.
Also:
* Implemented a method for populated message boxes when a JSON API request fails. So if a Save fails the user is now told why.
* Finished implementing EDBasecamp Model/ Basecamp repository
* Finished supplying methods for accessing associated models from any given model. For example from a survey, basecamp or world survey you get request the world that the model belongs to. From a world you can access list all assoicated surveys. From a basecamp you can list all basecamp surveys.

And...
* Followed @klightspeed's suggestion on a better Payload construction. Thanks for that! :)

### DIFF
--- a/EDDiscovery/EDDiscoveryForm.cs
+++ b/EDDiscovery/EDDiscoveryForm.cs
@@ -216,16 +216,16 @@ namespace EDDiscovery
 
                 CheckForNewInstaller();
 
-#if DEBUG
-//Robby : not sure about this, should we be moaning about this, what if people don't want to use EDmaterializer.
+                // Note: This is not for the end user, it's actually an alarm for us so we 
+                // know the file is missing from the installer. Without this credentials file 
+                // Prospecting data writes are broken.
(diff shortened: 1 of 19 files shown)


EDDiscovery/EDDiscoveryForm.cs
```diff
@@ -216,16 +216,16 @@ private void EDDiscoveryForm_Shown(object sender, EventArgs e)
 
                 CheckForNewInstaller();
 
-#if DEBUG
-//Robby : not sure about this, should we be moaning about this, what if people don't want to use EDmaterializer.
+                // Note: This is not for the end user, it's actually an alarm for us so we 
+                // know the file is missing from the installer. Without this credentials file 
+                // Prospecting data writes are broken.
```
gregmalcolm: @robbyxp1 Answered your comment question here... :)

PR #233, opened 2016-05-13
Minor code readability fixes

julienadam:
Hey guys,

I'm getting acquainted with the codebase so I took the opportunity to extract a few things and add some unit tests for them, introduce constants to get rid of magic numbers, remove unused event handlers etc.

I also added a dependency on NFluent to the unit tests project. It's a cool library that provides more expressive assertions and nicer assertion failure messages.

I realize reading the diff that I've used C# 6 features like auto-property initializers and string interpolation out of habit. Let me know if you prefer not to do that.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ packages
 /Installer/EDDiscovery
 */*Secrets.config
 /EDDiscovery.v12.suo
+*.ncrunch*
--- a/EDDiscovery/Controls/DataGridViewExtensions.cs
+++ b/EDDiscovery/Controls/DataGridViewExtensions.cs
@@ -1,0 +1,24 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Windows.Forms;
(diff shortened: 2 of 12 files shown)


gregmalcolm:
Oh wow, more tests. Nice! :)

julienadam:
I don't know about Travis, but I've used Appveyor in the past for that and it works nicely. .NET is supported out of the box, basically just point to a github repo with a .sln file somewhere and it does the rest.

gregmalcolm:
That could be a good way to do it because I think Travis only works for
.NET Mono projects. Assuming the stackoverflow post I saw is up to date.

julienadam:
I've added a build on appveyor here : https://ci.appveyor.com/project/julienadam/eddiscovery. Currently red because one of the unit tests in the master branch isn't passing.

finwen:
Nice refactoring.  Was just little work to merge it into ,y 3.2 branch that i had modified on the same places.  But en result seems very nice now :)

PR #238, opened 2016-05-19
Update 3D Map visited systems when commander is changed

klightspeed:
This should fix the bug where the visited systems are not updated when the user switches commanders.


robbyxp1:
oh dear. I fixed that one on my branch.. i hope we are not clashing

PR #237, opened 2016-05-17
Move tests to NUnit

klightspeed:
This, along with an appropriate `.travis.yml` and `appveyor.yml` will allow build tests on both Windows and Mono.  It also fixes some tests to match the changes that have been made in 3.2.

Do we want to re-add unit-tests to the solution, and add `.travis.yml` and `appveyor.yml` settings files?

Travis CI: [travis](https://travis-ci.org/klightspeed/EDDiscovery)
Appveyor: [appveyor](https://ci.appveyor.com/project/klightspeed/eddiscovery)

PR #239, opened 2016-05-19
Move tests to NUnit [rebased onto master]; Add CI settings

klightspeed:
This rebases the NUnit tests (from #237) onto master, and adds CI build settings for AppVeyor and Travis-CI.

This is based on discussion on #237.

All tests pass under AppVeyor and under Visual Studio.

Under Travis-CI, the `DatasetBuilder` tests will fail until the Mono SQLite fixes are merged.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,10 @@
+language: csharp
+solution: EDDiscovery.sln
+mono:
+  - latest
+install:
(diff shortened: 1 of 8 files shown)


klightspeed:
Note that if you want the AppVeyor build version to match the product version, the `version:` line will need to be updated when the version is updated.

PR #240, opened 2016-05-20
Betterstarnaming

robbyxp1:
Please pull into 3.2 or master.


EDDiscovery/3DMap/FormMap.cs
```diff
@@ -277,6 +279,8 @@ private ToolStripButton _AddExpedition(string name, Func<DateTime> starttime, Fu
 
         private void FillExpeditions()
         {
+            if (VisitedSystems == null || VisitedSystems.Count < 1)
```
klightspeed: This means that if the visited system history is empty when the map is first opened, the history filter list will be empty until the map is reloaded.

PR #243, opened 2016-05-23
Fetch system notes before systems

klightspeed:
This should resolve the system notes being missing from the travel history.


mwerle:
Confirmed works. (Conversation about issue was in Discord)

PR #252, opened 2016-06-23
Merge master into Mono-sqlite

klightspeed:


klightspeed:
Need to fix a few compilation errors, which I will add to this PR.

PR #264, opened 2016-07-15
Mono compatibility fixes

klightspeed:
This includes fixes for both Mono and Windows

* Use Mono.Data.Sqlite when System.Data.SQLite is non-functional (e.g. under Mono)
* Allow using default Windows themes
* Use a Status Strip for resizing (can also be used for status)
* Use a borderless window instead of a form element for the themed combo box dropdown
* Fix Toolbar styling under Mono
* Mono doesn't honour NCHITTEST, so move the window ourselves if it's not honoured
* Make SQLiteDBClass a static class, since we don't use any instance methods in it

Most of the commits are merges and fixes tracking the master and then RemoveGlobalSystems branches over the last 2 months.


klightspeed:
Just putting this pull request in here to track mergability and so we don't forget this.

klightspeed:
I will rebase this, culling all of the useless merge commits in the Mono-sqlite branch.

PR #265, opened 2016-07-15
UI behaviour improvements for 3.3

klightspeed:
* Use a colour table for toolstrips etc.  This should reduce the amount of custom rendering code for toolstrips, improving compatibility.
* Replace the drawn panel sizing grip with the status strip's sizing grip.  This also allows us to anchor the bottom of the tab panels to the top of the status strip.
* Rework the custom combo box, using a borderless window instead of a floating panel on the main form.


finwen:
waiting with this until 4.0.x is stabel.

PR #269, opened 2016-07-17
Use EDDB/v4/systems_populated.json

klightspeed:
Download the 7MB filtered EDDB systems json instead of the 300MB unfiltered EDDB systems json


klightspeed:
Duplicate of 2c2b458

Issue #267, opened 2016-07-16
Make the matching better in System Class between VSC and Db


robbyxp1:
FillVisitedSystems... 

search for name
if select returns more than 1
    pick the entry which has the best x/y/z to the VSC x/y/z
    if VSC does not have a x/y/z or x/y/z does not match
         mark as ambiguous in memory
    show on screen somehow its ambiguous

robbyxp1:
VisitedSystem: FillVisitedSystems

VisitedSystems has a in memory only entry - alt_name (or something like that)

search for name
    if 1 name match
        if VSC has x/y/z
            if VSC x/y/z != db x/y/z
                search for star at x/y/z
                    if match
                    {
                        vsc.alt_name = "(" vsc.cur_name ")"
                        vcs.cur_name = eddb.name
                        update database
                    }
            else
                star found
        else
            star found


    if >1 name match
        if VSC has x/y/z
            search for best x/y/z

            if match x/y/z
                vsc.alt_name = "(" vsc.cur_name ")"
                vcs.cur_name = eddb.name
                update database
            else
                vsc.alt_name = "(Not Found)"

        else
            vcs.alt_name = "(Ambiguous)"


Right Click - Rename
    allow a rename

Right Click - find alternates
    if VSC has x/y/z
        present list of stars at x/y/z and allow a rename

klightspeed:
If the VisitedSystem is newer than the db entry, shouldn't the vsc name override the db name?

robbyxp1:
good point. Can't decide.  here is a better logic but without an auto rename.

VisitedSystem: FillVisitedSystems

VisitedSystems has a new in memory only entry - alt_name - tells you the results.. user decides what to do.

in memory since we want to do this every time in case the db has more stars..

search for name
    if 0 match
        if VSC has x/y/z
                search for star at x/y/z
                    if match
                    {
                        vsc.alt_name = "EDSM(" db name ")"
                    }

    if 1 name match
        if VSC has x/y/z
            if VSC x/y/z != db x/y/z
                search for star at x/y/z
                    if match
                        vsc.alt_name = "EDSM(" db name ")"

                    no star match (so the dummy is created at the right position)
            else
                star found
        else
            star found


    if >1 name match
        if VSC has x/y/z
            search for best x/y/z

            if match x/y/z
                vsc.alt_name = "EDSM(" db name ")"

            no star match (so the dummy is created at the right position)

        else
[...]

PR #270, opened 2016-07-17
Match systems by name, position and/or EDSM ID

klightspeed:
This should match systems as discussed in #267 
This still needs to be hooked into the UI.


klightspeed:
Moving branch to main repo

PR #278, opened 2016-07-23
Mono compatibility fixes

klightspeed:
* Make SQLiteDBClass a static class, as nothing uses any instances of it
* Use a `GetSqliteProviderFactory()` method to get the provider factory
* Have this factory provide a `Mono.Sqlite.SqliteProviderFactory` when System.Data.SQLite is non-functional (e.g. under Mono, or perhaps on ARM)
* Merge the DbConnection etc. extension methods into the SQLiteDBClass static class
* Mono does not honour the results of NCHITTEST, so move the window ourselves if a mousedown / mousemove is detected where we said the non-client area is
* Allow the user to select that they don't want to use a theme - i.e. use the system theme.

### DIFF
--- a/EDDiscovery/DB/SQLiteDBClass.cs
+++ b/EDDiscovery/DB/SQLiteDBClass.cs
@@ -368,7 +368,13 @@ namespace EDDiscovery.DB
 
     public static class SQLiteDBClass
     {
+        #region Private properties / fields
         private static DbProviderFactory _factory;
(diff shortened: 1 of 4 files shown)


klightspeed:
This contains the meat of the changes from the Mono-sqlite branch, rebased onto the current master branch.

Probably test that everything else works before merging this in.

PR #294, opened 2016-07-27
Re-organise netlog parsing code

klightspeed:
This is a significant re-organisation of the netlog reader code.

* Use a reader that will only return full lines, and keeps the offset within the file of the last byte of the last line it has returned
* Replace the NetLogFileInfo structures with this reader
* Keep the time of the last log entry read in the reader
* Separate the code that finds and returns a system line into a separate function
* Add a timezone-aware header parsing function
* Parse and process each system as it is read from the logs
* Instead of testing if the file reports a larger size, simply try to read more of the file on each tick.
* In ParseFiles, compare the system with that immediately before it rather than the last one in the list.

It is hoped that the base reader can also be used for reading the journal files (perhaps separated into a separate class file and subclassing that for the netlog reader code).

### DIFF
--- a/EDDiscovery/DB/VisitedSystemsClass.cs
+++ b/EDDiscovery/DB/VisitedSystemsClass.cs
@@ -64,6 +64,36 @@ namespace EDDiscovery2.DB
             }
         }
 
+        public VisitedSystemsClass(DbDataReader reader)
+        {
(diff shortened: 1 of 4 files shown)


robbyxp1:
Lets pause on this for a moment and not put it in 4.1

PR #304, opened 2016-07-31
Automatically upgrade DB to latest schema

klightspeed:
This should enable automatically upgrading the database schema to the latest schema.

It does not currently make any changes that would drop columns or change column types, as these would require dropping and re-creating the table as show in the SQLite [Making Other Kinds Of Table Schema Changes](https://www.sqlite.org/lang_altertable.html#otheralter) section of the `ALTER TABLE` page.

This still needs to be tested for upgrading from various versions of the database schema.


klightspeed:
This has been merged into v5-database-split.

finwen:
Interesting idea. Will test next week.

finwen:
Save this good idea for EDD V5 i think.

Issue #313, opened 2016-08-03
Portable Install


mwerle:
Request to have a portable install of EDDiscovery.

"Portable" in this sense means that the app is installed standalone in a directory, with -all- data and configuration files kept in the installation directory. No registry entries, no files in %APPDATA% or %USERPROFILE%, etc.

https://forums.frontier.co.uk/showthread.php/138155-EDDiscovery?p=4302209&viewfull=1#post4302209

robbyxp1:
Before anyone does anything, we already have a -appfolder option to move the ED folder to another one in appdata\local.  This probably needs an file in the .exe to tell it.. low priority

klightspeed:
This shouldn't be too difficult - just add the possibility for an option in `EDDiscovery.exe.config`

Red13x3:
+ > 9,000!

Should have been from the get-go. 

--
Mark
iPhone: 0402 322 901

robbyxp1:
How is it going to be packaged.. does the installer offer this option? etc. More to it than meets the eye.

mwerle:
I just added this as a backlog item as it was requested in the forum; I didn't intend anybody to drop everything and start working on it immediately. I agree - low priority.

klightspeed:
Zip up the contents of the EDDiscovery/bin/Debug directory, or add a MSBuild project that does that for you?

PR #314, opened 2016-08-05
Create EDDiscovery.Portable.zip as part of the build process

klightspeed:
Requested in #313


robbyxp1:
This one is for Finwen

PR #340, opened 2016-08-11
Fix a couple of races in the 3D map; add per-star colouring

klightspeed:
* Fix a race where a dataset being updated by the background could be referenced by the foreground while it is set to null
* Fix a race where a dataset could be updated between check and use.
* Add per-star colouring, currently based on `RandomId`


klightspeed:
@robbyxp1 This is for after you have pushed your changes.  If there are any merge conflicts, I will fix them.

PR #370, opened 2016-08-29
Tweaked trilateration control margins to stop it dropping off the bottom of the form

iainross:
(I have no idea why Visual Studio decided to rearrange the rest of the file in a seemingly arbitrary fashion though....)


klightspeed:
The random musical chairs Visual Studio plays is why I usually take a diff, cut out the musical chairs parts, then re-apply the diff.

PR #393, opened 2016-09-07
Merge fixes from 4.3 into master

klightspeed:
* Only log long running command once per transaction
* Fix the model view matrix for orthogonal view


klightspeed:
I'll create a different pull request based on #392

PR #407, opened 2016-09-09
Fixes to ensure CommanderList is never empty; specifically:

mwerle:
* fix creating a new commander which requires a non-null EdsmApiKey.
* fix reading values from DB (some conversion were throwing exceptions due to long -> int or NULL -> string)

Rename "NetLogPath" -> "NetLogDir" throughout the code to ensure consistency and prevent future bugs (attempt was made to read the wrong one from the database).

### DIFF
--- a/EDDiscovery/EDDConfig.cs
+++ b/EDDiscovery/EDDConfig.cs
@@ -337,9 +337,9 @@ namespace EDDiscovery2
                     {
                         while (reader.Read())
                         {
-                            int id = (int)(long)reader["Id"];
-                            string name = (string)reader["Name"];
-                            string edsmapikey = (string)reader["EdsmApiKey"];
+                            int id = Convert.ToInt32(reader["Id"]);
+                            string name = Convert.ToString(reader["Name"]);
+                            string edsmapikey = Convert.ToString(reader["EdsmApiKey"]);
 
                             if (id > maxnr)
                                 maxnr = id;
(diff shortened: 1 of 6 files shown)


EDDiscovery/EDDConfig.cs
```diff
@@ -373,15 +373,15 @@ private void LoadCommanders()
                             EDCommander cmdr = new EDCommander(0,
                                 SQLiteDBClass.GetSettingString("EDCommanderName" + i.ToString(), commanderName, conn),
                                 SQLiteDBClass.GetSettingString("EDCommanderApiKey" + i.ToString(), apikey, conn));
-                            cmdr.NetLogPath = SQLiteDBClass.GetSettingString("EDCommanderNetLogPath" + i.ToString(), null, conn);
+                            cmdr.NetLogDir = SQLiteDBClass.GetSettingString("EDCommanderNetLogDir" + i.ToString(), null, conn);
```
klightspeed: This will cause it to fail to migrate the commander's Netlog directory

EDDiscovery/EDDConfig.cs
```diff
@@ -373,15 +373,15 @@ private void LoadCommanders()
                             EDCommander cmdr = new EDCommander(0,
```
klightspeed: Replace `0` with `i`

EDDiscovery/EDDConfig.cs
mwerle: Argh, I checked 3 times; I thought I was doing the rename the right-way around to keep the DB setting the same. Or did I rename one item I shouldn't have?

PR #408, opened 2016-09-09
Fix commander netlog path migration

klightspeed:


klightspeed:
@finwen pushed the same thing.

PR #422, opened 2016-09-13
Use Journal Reader commander if possible during import

klightspeed:


finwen:
with that solution the 2 first  ntries will be mapped to wrong commander if  we are viewing another commander and parsing a new file.(ändrad)
if we sett commander iun the travellog unit  we can  get right commander from tlu instead?
reader.Commander is set we could use it

klightspeed:
I will redirect this pull request to the EDJournal branch.

PR #427, opened 2016-09-13
Fix an apparent race in ToggleSummaryPopOut() [4.3]

klightspeed:
It looks almost as if Windows will occasionally sneak
a callback in while processing another window message.

For example, #425 appears to have been from Windows
sneaking in the RefreshWorkerComplete callback after
calling WM_CLOSE on the summaryPopOut, but before returning
from summaryPopOut.Close()


robbyxp1:
It does not deal with all issues, if the user closes the spanel using the taskbar, it does not come back on the next click.

PR #449, opened 2016-09-24
Remove all references to VisitedSystemsClass

klightspeed:
* Return JournalLocOrJump entries from NetLogClass
* Comment out any other (now unused) references to VisitedSystemsClass
* Remove VisitedSystemsClass from the compilation sources
* Add helper function to allow parsing logs from an arbitrary directory


klightspeed:
One moment - need to fix some stuff I missed.

klightspeed:
Should be good now.

PR #465, opened 2016-09-28
Filter journal by commander

klightspeed:


robbyxp1:
This will clash

robbyxp1:
Sorry consider

robbyxp1:
Right, i've basically taken your structure and moved it to the next PR, so the merge can be clean.

PR #471, opened 2016-09-30
Use the fast modtime checking if dumps ordered by date

klightspeed:
The beta.edsm.net system dumps are currently still ordered by X coordinate ascending, so the assumption that the dump is ordered by date (as it is on www.edsm.net) does not hold there.

Check if the dates in the dump are ascending, and if not then use `GetLastSystemModifiedTime()` which would require another expensive index to speed it up.


klightspeed:
One moment.

klightspeed:
Should be good now.

Issue #499, opened 2016-10-25
Export buttin on exporttab not visible with some themes


finwen:
With themes with big fonts the Export button will be hidden under the groupbox.

finwen:
Solves #49

PR #501, opened 2016-10-26
Make Export tab layout dynamic

klightspeed:
This should prevent e.g. the Export button from being hidden by the groupbox to its left.


finwen:
Solves #499

PR #518, opened 2016-10-27
Add "EdsmName" column to Commanders table

klightspeed:
Also add a `ShortName` column to the `MaterialsCommodities` table for @robbyxp1


klightspeed:
One moment...

PR #538, opened 2016-10-31
Prevent DB deadlocks

klightspeed:
It appears that the new SQLite library uses speculative
locking - it locks the database when it is freshly
opened, in anticipation of a command being executed.
Prevent the deadlocks this creates by locking the database
for the entire life of short-lived connections.

Also remove some recursive DB opens, and report any other
recursive opens.


klightspeed:
Superseded by #539

PR #596, opened 2016-11-30
Include resurrect locations on map and EDSM sync

### LINKED ISSUE #560: Bug: When Dying the respawn System doesn't count as visited System
I did die yesterday. But the respawn System doesn't show in my log as visited at that point in time since the respawn event isn't counted as a travel event.

This results in the 3D map as a big jump from the System I died in to the first System I jumped after respawn. (As a result to that also not send to EDSM)

Version 5.0.17

klightspeed:
* Include initial location and location after death / resurrect on map and in EDSM sync
* Split travel line on death / resurrect

This should fix #560

### DIFF
--- a/EDDiscovery/3DMap/DatasetBuilder.cs
+++ b/EDDiscovery/3DMap/DatasetBuilder.cs
@@ -103,7 +103,7 @@ namespace EDDiscovery2._3DMap
 
             if (syslists != null)
             {
-                foreach (HistoryEntry vs in syslists)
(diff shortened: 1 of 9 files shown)


AnthorNet:
Something I want to add to our logs API, death or resurrect? 
Which one seems better for you with the journal event to be associated with a log API call?

klightspeed:
A `Died` event can occur at any time while in a system, while a `Resurrect` event should always be associated with the `Docked` and `Location` events that follow it.

PR #599, opened 2016-12-02
Basic idea for a fuel warning

jthorpe4:
Let me know what you think, warning is hard coded to 25%


EDDiscovery/HistoryList.cs
```diff
@@ -70,6 +71,11 @@ public bool ISEDDNMessage
         public int TravelledMissingjump { get { return travelled_missingjump; } }
         int travelled_jumps;
 
+        private static double last_fuelscooptotal = 0.0;
```
klightspeed: Is this meant to be static?

jthorpe4:
Yes, even if I wasn't happy about it. I wanted to temporary hold the last
value for total fuel level found in fuel scoop. I guess thinking about it I
could pull the last fuel scoop even from history list.

EDDiscovery/HistoryList.cs
jthorpe4: I had a quick look at this again this morning and I can change the code to avoid the static, I'll have go tonight.

PR #603, opened 2016-12-03
Refactor DB initialization

klightspeed:
* Merge `SQLiteDBClassUser` into `SQLiteConnectionUser` and move it to its own source file
* Merge `SQLiteDBClassSystem` into `SQLiteConnectionSystem` and move it to its own source file
* Merge the DB factory creation and DB open into SQLiteConnectionED
* Cache the register on startup and use the cache during initialization
* Cache the commanders on startup
* Move saved routes load and compute stars thread startup to after DB init
* Initialize DB in background, and show splash window on foreground thread
* Don't show splash unless DB init is still running on form load


klightspeed:
Do we want this in e.g. 5.2?

Issue #598, opened 2016-11-30
Ledger Sell events not adding to balance correctly


svensven:
On a Market Sell event, the ledger is adding 1 unit's sale price to the balance, rather than number-of-units * sale price.
E.g.:
![eddiscovery ledger](https://cloud.githubusercontent.com/assets/5736616/20773772/6c96a9b8-b74b-11e6-98c4-42fe94cc6aff.PNG)

svensven:
At a guess, the 5th parameter on line 50 of EliteDangerous/JournalEvents/JournalMarketSell.cs
should be TotalSale rather than SellPrice?

eproces:
Having same issue, attempting a fix now.

svensven:
Fixed in 5.1.6.0. Thanks :)

Issue #605, opened 2016-12-06
EDD 5.1.5 - Ledger not deducting donation missions from balance


CMDRLucienn:
EDDiscovery v5.1.5 does not appear to be deducting donation missions from the running balance. It only corrects itself on the Load Game event:

![image](https://cloud.githubusercontent.com/assets/7802122/20946121/d0317d2a-bbc6-11e6-93bc-0df8519e0a58.png)

It seems that EDDiscovery is not reading the "Donation":XXXXXX field of the Mission Complete event.

svensven:
Confirming fixed in 5.1.6.0.

PR #607, opened 2016-12-08
Update Journal Ledger Entries

eproces:
Correctly pushes (unit price x units) to the ledger Credit column, corrects ledger balance issue.

Fixes issue #598.


eproces:
Added a commit to also fix issue #605. 

Donation was already there, so by making sure Donation and Reward are their correct value or zero, just subtract and you get the right debit or credit without an if statement.

finwen:
Looks good will test tonight.

Issue #602, opened 2016-12-02
Feature Request - Habitable zone calculation


kjnpbr:
Any chance of after scanning a star that the habitable zone distances could be calculated and displayed?

finwen:
Depends if someone can define a calculation on habitable zone fir different star types. 


The estimation on habitable zone for out solar system varies lots between different theories...

finwen:
Will be in release 5.2

PR #608, opened 2016-12-10
Add Hab. zone approximation to display info for stars

iainross:
Calculation lifted from Jackie Silver's hab zone calculator with their permission.


finwen:
Solves #602

PR #628, opened 2016-12-15
Remove OpenTK reference from EDDiscoveryTests project

klightspeed:
OpenTK isn't used by the test project, and it was sitting at an old version, causing a reference mismatch warning on build.


klightspeed:
Bah - the namespace is apparently referenced.

PR #671, opened 2017-01-06
Route planning output textbox should be read-only

phroggster:
Hopefully, I'll eventually quit trying to keyboard steer my ship while I have my next destination highlighted for copying. Barring that, I should at least not have to worry about overwriting the text telling me where to
go next.

Note that my C# 6+ compiler is presently out of order. You'll want to compile and test this before actually applying this change. Visual effects may also make it less legible, but I never did really care what usability experts thought.


phroggster:
Nevermind. I clearly need to get my compiler operational before I try this again.

PR #672, opened 2017-01-06
Routing information should be read-only

phroggster:
If only I could learn to stop trying to pilot my ship while I've got my next destination selected for copying in EDD. At least with this patch, the destination won't be aaadadddadaddadadwj.

[RichTextBoxScroll](https://github.com/EDDiscovery/EDDiscovery/blob/master/EDDiscovery/Controls/RichTextBoxScroll.cs#L37) has the following comment that may impact this:
`public RichTextBoxBack TextBox;  // Use these with caution`
Not sure where a problem could arise from using this, but I wanted to note it here. I'm also not very familiar with the EDD code base, so YMMV.

### DIFF
--- a/EDDiscovery/RouteControl.cs
+++ b/EDDiscovery/RouteControl.cs
@@ -44,6 +44,7 @@ namespace EDDiscovery
             InitializeComponent();
             button_Route.Enabled = false;
             cmd3DMap.Enabled = false;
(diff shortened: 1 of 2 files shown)


phroggster:
I noticed that the log controls had the same issue, so here's a functionally identical tweak for those.

Issue #417, opened 2016-09-13
Feature Request


MalikDama:
Options to minimize to System Tray

this would be nice for  an alt-tab-aholic

phroggster:
The reference to 098d426c77e87a1cbc09f58d70e105e9769b06b0 was unintentional. For some reason I typed in this ticket instead of #713. Apologies for the spam.

PR #673, opened 2017-01-06
Add proper multi-instance handling

phroggster:
Win32 APIs are horrible, but sometimes, they're the only way to get work done. Application will now attempt to focus the original window if it gets launched twice. If we can't raise/focus that window, fallback to the good old messagebox prompt.

Feedback welcome.

### DIFF
--- a/EDDiscovery/Program.cs
+++ b/EDDiscovery/Program.cs
@@ -1,8 +1,9 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Diagnostics;               //Process
 using System.Linq;
 using System.Reflection;                //Assembly
-using System.Runtime.InteropServices;   //GuidAttribute
+using System.Runtime.InteropServices;   //GuidAttribute, Win32 DllImports: SetForegoundWindow, ShowWindowAsync, IsIconic
 using System.Security.AccessControl;    //MutexAccessRule
 using System.Security.Principal;        //SecurityIdentifier
 using System.Threading;                 // Tasks and Mutex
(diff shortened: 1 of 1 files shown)


EDDiscovery/Program.cs
```diff
@@ -31,15 +63,40 @@ static void Main()
...
+        /// <returns><c>true</c> if the other application windw was raised. <c>false</c> otherwise</returns>
+        static bool RaiseOtherProcess()
+        {
+            bool retval = false;
+            Process proc = Process.GetCurrentProcess();
+            foreach (Process otherProc in Process.GetProcessesByName(proc.ProcessName))
+            {
+                if (proc.Id != otherProc.Id)
+                {
+                    IntPtr hWnd = otherProc.MainWindowHandle;
```
klightspeed: A portable alternative might be:
```
Microsoft.VisualBasic.Interaction.AppActivate(otherProc.Id);
```
Though that won't restore the app if it's minimized.

This block could be wrapped in `try { } catch (DllNotFoundException) { } catch { }`, where the portable function is called if binding the Win32 functions fail, and `retval` is set to `false` if any other exception is caught.

finwen:
Will test it litte later and want to know  klightspeed's thought about it with Mono build.

phroggster:
This'll make Mono crash and burn faster than if it was being driven by a Kerbal. I'll submit some fixes in short order, soon as I can figure out whether #417 will need any of this.

finwen:
Good.  I think i wait to merge this until we start working on 5.3 (soon)

phroggster:
Sure enough, with the feature417 changeset applied, this approach falls to garbage. Going to have to go at the IPC from a different angle.

klightspeed:
You would probably need something like a named pipe (`System.IO.Pipes.NamedPipeServerStream` / `System.IO.Pipes.NamedPipeClientStream`, which should be cross-platform) and/or a custom window message (processed in `EDDiscovery.EDDiscoveryForm.WindowProc`, which will only work on Windows).

phroggster:
Closing this PR to get it out of the way. Bigger (and much better) solution coming in short order.

PR #692, opened 2017-01-10
Fix a NRE when loading an empty route

phroggster:
Fix the NRE noted by [Uri (Mila strelok)](https://cdn.discordapp.com/attachments/151028492048269313/268480665098518539/unknown.png)


phroggster:
The second commit fixes that bug when right-clicking an unknown system (also found by [Uri (Mila strelok)](https://cdn.discordapp.com/attachments/151028492048269313/268485615920611329/unknown.png)

PR #696, opened 2017-01-12
Fixed what looked like a few rounding issues

jthorpe4:
I seem to have a 0.5T extra which i believe might be my SRV but I need to dock to confirm that


klightspeed:
Duckfather on the distantstars channel in FleetComm mentioned something about the jump range on their Anaconda changing after 2.2.03 due to an increased maximum mass reduction on enhanced low power shields.

I think the SRV was 4T with 0.5T fuel.

jthorpe4:
I've just added an offset of 0.5T in my current cargo, it will do me for now, it might be a enginnering mod, i think i have one on power plant and thrusters but i can't remember what ones I need to check once docked.   
The Jump range pretty closely matches the in-game one now, i'm just sometimes 0.01LY out and I think that's got to be another round issue somewhere.

PR #700, opened 2017-01-13
Fix #699 (import saved routes with unknown systems)

### LINKED ISSUE #699: Import Route WITH unknown systems
Simple enough, when I import a route from a txt file, I want an option to import it with any unknown system in the list, currently you can only choose between not importing the whole file, or import it without the unknown systems, and I want it to import EVERYTHING.

9f5220fe7c 1

PS: Spelling error in the message too ;)

phroggster:
Let the DataGridView process the route instead of trying to address it during import. This has the added benefit of letting the user know where along the route the error appears, as it will be highlighted and lack
distance information.

### DIFF
--- a/EDDiscovery/SavedRouteExpeditionControl.cs
+++ b/EDDiscovery/SavedRouteExpeditionControl.cs
@@ -801,13 +801,12 @@ namespace EDDiscovery
             }
             catch (IOException)
             {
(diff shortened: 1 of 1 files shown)


phroggster:
Note that #701 should be applied in tandem with this, to address my oversight in the route tracker s-panel.

PR #706, opened 2017-01-15
Basic autoscrolling to keep the most recent history steps in view

jeoffman:
I made a corresponding "issue" #705 - keep most recent travel data in view (scroll grid to top)

PR #710, opened 2017-01-15
autoscroll Journal based on autoscroll selection from History

jeoffman:
using feedback from previous pull request = Issue #706 

Thanks to klightspeed and Lee for your help!


finwen:
Should we kill your old PR #706 then?

PR #711, opened 2017-01-16
Added support for Jet cone boosts

jthorpe4:
Maybe need to add support for double stacking boost jumps but that's probably an exploit that will get nerfed.


klightspeed:
The boosts do not stack, but instead the second boost carries through to the next system.

jthorpe4:
Yes, that's what I meant :) sorry.  either way it's probably an exploit, even if its quite cool

PR #712, opened 2017-01-16
Separate the UI from the logic in EDDiscoveryForm

klightspeed:
This should allow the logic to be plugged into a different UI (such as a macOS UI)


finwen:
Waiting with this about a week.  until robbys branch is more ready so i can test merge both of them,

klightspeed:
I'll work on rebasing this onto master, to get rid of all of the merges in this PR.

PR #717, opened 2017-01-16
Don't save window size if minimized

phroggster:
This should hopefully address the unusable small windows that require -NoRepositionWindow or database hacking.

For example, see [this image](http://i.imgur.com/LEEFXB7.jpg).

### DIFF
--- a/EDDiscovery/EDDiscoveryForm.Designer.cs
+++ b/EDDiscovery/EDDiscoveryForm.Designer.cs
@@ -788,6 +788,7 @@ namespace EDDiscovery
             this.FormClosing += new System.Windows.Forms.FormClosingEventHandler(this.EDDiscoveryForm_FormClosing);
             this.Load += new System.EventHandler(this.EDDiscoveryForm_Load);
             this.Resize += new System.EventHandler(this.EDDiscoveryForm_Resize);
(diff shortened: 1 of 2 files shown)


phroggster:
Playing around after this patch, at least in Windows 10, exiting while minimized will restore the window size appropriately but the position will be reset to the far top-left. Exiting while maximized or normal will still result in appropriate behavior.

klightspeed:
Perhaps we should remember the window size and position immediately before being minimized or maximized (i.e. store the window position and size in `OnResizeEnd` and/or `OnResize` if the window state is `Normal`

phroggster:
Give that third revision a shot, see if it behaves any better.

klightspeed:
`Form.Resize` is not triggered when the window is moved - you would need to attach a handler to the `ResizeEnd` event to capture the size and position after the window is moved.

phroggster:
The fourth time is a charm?

phroggster:
Ugh, _formMax shouldn't be stored if minimized. Any other issues visible to you klightspeed?

klightspeed:
Should work.

phroggster:
Ok, just pushed a correction for `_formMax`, this PR should be ready for prime time.

PR #736, opened 2017-01-21
Extract system name from Screenshot event

klightspeed:
This takes advantage of the `Screenshot` journal event to link a screenshot to a system.

It will delay the screenshot conversion by up to 5 seconds if it's a bmp screenshot and no associated Screenshot event is written to the journal.


robbyxp1:
Why is this better than the current system?

PR #742, opened 2017-01-23
Display planetary coordinates in degrees decimal

phroggster:
... instead of Degrees Minutes Seconds. This makes sharing and revisiting landing and POI locations easier, as it saves users from converting DMS into degrees decimal.

- Before: `Latitude:34°S 12'57", Longitude:77°W 41'34"`
- After:   `Latitude:-34.2159° (S), Longitude:-77.6928° (W)`

I'm open for comments and suggestions, especially with regards to the non-standard use of the `{N,E,S,W}` indicators and/or the non-standard use of the `°` symbol. If ED ever defaults to (or even supports) display of coordinates in DMS, I believe that this change should be reverted.

### DIFF
--- a/EDDiscovery/JSON/JSONPrettyPrint.cs
+++ b/EDDiscovery/JSON/JSONPrettyPrint.cs
@@ -153,7 +153,7 @@ namespace EDDiscovery
                             {
                                 string marker = (lv >= 0.0d) ? "(N)" : "(S)";   // presume lat
                                 if (converters[i].converttype == Types.TLong)
-                                    marker = (lv >= 0) ? "(E)" : "(W)";         // but handle long
+                                    marker = (lv >= 0.0d) ? "(E)" : "(W)";      // but handle long
                                 value = string.Format("{0:0.0000}° {1}", lv, marker);
                                 if (formatsplit.Length >= 2)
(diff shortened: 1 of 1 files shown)


lausianne:
Just to show support: I totally agree with this. I use a Google Spreadheet where I enter my current location and the target location and it gives me bearing and distance. Coordinates are all in decimals. Whenever coordinates are mentioned anywhere (forums, Reddit), it's always decimals.
Dream feature: EDDiscovery allows me to enter target coordinates and shows me where to go (to replace my spreadsheet ...)

phroggster:
> we need to add `lv = Math.Abs(lv);`

I explicitly avoided such so that we would show the same coordinates as displayed in the game and utilized by other third-party tools. If we want to remove the (potentially confusing, but potentially helpful) NESW indicators, that's fine, but I believe that displaying the negative sign is important.

IMHO, the only time planetary coordinates are ever used is to revisit a POI, to inform others of where a POI is, or to calculate a bearing towards a POI. In all three cases, signed degrees decimal are the expected format.

phroggster:
The consensus that was reached was that we should utilize settings for units. Will resubmit with enhancements.

PR #746, opened 2017-01-24
My apologies to Anthor (master)

phroggster:
The notification icon context menu "Sync with EDSM" got hooked up to the wrong method. Instead of syncing recent events, it was performing a full sync.

Added some commentary for the notifyicon events.


finwen:
The bigger question are should b we have Sync menu items at all in notify area.  The menu items should be used rarely so it could be moved out from notify icon i think.

phroggster:
Will gut the sync ContextMenuItems and resubmit.

phroggster:
This went into 164f5e74847ccbbd431d629849393f22189c0184, whereby we simply remove the sync functionality from the context menu.

PR #806, opened 2017-02-06
Add AutoCompleteDGV* controls to ExtendedControls

phroggster:
This inherits vastly from AutoCompleteTextBox and still has some rough.edges, but is mostly usable.
Wired SavedRouteExpeditionControl up to use this new control.
*Enhanced* EDDTheme to allow for single-instancing, similar to EDDConfig.

Eyeballs will be appreciated here, particularly with EDDTheme change (and updating *some* things to use it if appropriate).


phroggster:
Unable to hook this into `UserControlExploration` as easily as I should have been. Retracting PR, will continue work and resubmit.

PR #811, opened 2017-02-07
Add a couple platform checks

phroggster:
@robbyxp1 Eyeballs on the speech synthesis change. This should allow for compile-time detection as well as run-time detection. I also inverted the `#if` to match the exact format that was used around the class definition.

Speech synthesis is Windows 2000+. Recognition will require Vista+.
	http://stackoverflow.com/a/5861145

AppendMenu, GetSystemMenu are also 2000+.
	https://msdn.microsoft.com/en-us/library/windows/desktop/ms647616(v=vs.85).aspx
	https://msdn.microsoft.com/en-us/library/windows/desktop/ms647985(v=vs.85).aspx

### DIFF
--- a/EDDiscovery/Forms/UserControlForm.cs
+++ b/EDDiscovery/Forms/UserControlForm.cs
@@ -398,23 +398,28 @@ namespace EDDiscovery.Forms
         [DllImport("user32.dll", CharSet = CharSet.Auto, SetLastError = true)]
         private static extern bool AppendMenu(IntPtr hMenu, int uFlags, int uIDNewItem, string lpNewItem);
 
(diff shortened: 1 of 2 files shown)


finwen:
We dont need this checks,  EDDiscovery is .Net 4.6  =>  Minimum Windows  Vista SP2.

phroggster:
Here's the [trace log](https://gist.github.com/anonymous/10f16974dfcccabf7ac5451c6929cdc8) for this, btw.

phroggster:
As discussed in Discord, the checks regarding version aren't really that important, but the `PlatformID.Win32NT` check will save some annoying popups on non-Windows hosts. The version checks are just safety frosting.

PR #816, opened 2017-02-08
Optimize a couple of expensive linq's

phroggster:
Parallelize GetVisitsCount and GetScans queries, and move the long comparisons ahead of the string comparisons. This vastly speeds up these queries, albeit at the cost of multi-core CPU utilization.

On my 6700k, it would take roughly 60 seconds for the previous queries to complete, which is past the timeout resulting in incorrect data display. Now, it's done in under 30ms, and the initially displayed value is accurate to boot. I encourage testing to see if this cuts too far into ED resources on various other processors.

One item to do: cache the previous system ID in `ShowSystemInformation()` and don't bother running the queries again if the same system was selected.

>Delta 2-7 - Today at 4:20 PM
Hi there I'm trying to troubleshoot an issue I have regarding systems visited I'm using EDDiscovery as well as EDDI with voiceattack (EDDI tells me on hyperjumping that this is the first time I'm visiting a system when it is local and I have been here dozens of times and when I cross check EDDiscovery it informs me in the Log the correct number but the Vissits box is locked at either 1 or 2 (Attached screenshot on this occassion shows 2 in visit box but total 5 in Log I'd like to try resolve this so records are correct but dont know best course of action as I'd hate to loose the data so far collected.

### DIFF
--- a/EDDiscovery/HistoryList.cs
+++ b/EDDiscovery/HistoryList.cs
@@ -646,11 +646,15 @@ namespace EDDiscovery
 
         public int GetVisitsCount(string name, long edsmid = 0)
         {
-            return historylist.Where(he => he.IsFSDJump && he.System.name.Equals(name, StringComparison.InvariantCultureIgnoreCase) && (edsmid <= 0 || he.System.id_edsm == edsmid)).Count();
+            return (from he in historylist.AsParallel()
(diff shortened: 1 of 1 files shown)


klightspeed:
Since a lot of stuff wants jumps or data associated with the current system, why don't we create a specialized SystemVisits list in HistoryList, which is filled in with just FSD Jumps and resurrections, each possibly with a list of all of the journal events associated with that system visit?

We could add a `SystemVisit` property to `HistoryEntry`, which points back at the associated system visit, and which would be filled out by the logic that associates the history entry to a system.

PR #820, opened 2017-02-11
Fix compilation errors in Mono; merge master into master-audioimprovements

klightspeed:


robbyxp1:
just hold this for a bit.. i won't be clashing with it, just want to make sure next bit works against stable base

klightspeed:
Will rebase.

PR #826, opened 2017-02-13
Calculate home distance from set home system

### LINKED ISSUE #823: TravelHistoryControl shows distance to Sol instead of distance to Home
I've changed the Home System on the 3D Map, which has worked for the Home button on the 3D Map but doesn't seem to translate back to the Top Right View of the History Page, which remains as Sol. I also can't find any other setting to change the Home System in the Top Right View of the History Page.

Am I missing something?

klightspeed:
Should fix #823

### DIFF
--- a/EDDiscovery/3DMap/FormMap.cs
+++ b/EDDiscovery/3DMap/FormMap.cs
@@ -135,10 +135,17 @@ namespace EDDiscovery2
         public void Prepare(ISystem historysel, string homesys, ISystem centersys, float zoom,
                             List<HistoryEntry> visited)
         {
+            ISystem homeSystem = (homesys != null) ? FindSystem(homesys) : null;
+            Prepare(historysel, homeSystem, centersys, zoom, visited);
+        }
+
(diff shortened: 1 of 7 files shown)


robbyxp1:
hmmm

robbyxp1:
thats a bit of history you removed there ;-)

iainross:
Oh noes!  That's going to mess me up - the home system is optionally used to centre the 3D Map, I want Sgr A* for that, but I want the Sol distance readily available on the front screen for filling in my ELW spreadsheet for merging with Marx's when I get home.

Hmm, I might have to add something extra here so I can keep those two things distinct.

phroggster:
There's enough space there for a distance to Sol and a distance to home textbox. Should probably even add "Dist." to the labels to make it clearer: {"Dist. Home", "Dist. Sol"}.

I also think that it should probably be using the tried and tested distance calculator in DB.SystemClass, or even add {`DistanceTo(ISystemBase)`, `DistanceTo(double, double, double)`} methods to ISystemBase for use when callers already know that their reference isn't null. No sense duplicating that code everywhere.

PR #831, opened 2017-02-15
Cleanup a font handle leak in DownloadManagerForm

phroggster:


robbyxp1:
i'll solve it in the current branch and merge back in thanks

phroggster:
@klightspeed pointed out that I was disposing of this in the wrong spot. I know you said that you were going to move this to `OnFormClosing()` or something, but if it stays in `Dispose(boolean)` make sure that said boolean is `true`, and that it must not be limited inside an `if (components != null)` check. Here's what it should *really* look like if *everything* is properly accounted for:

```CS
public void Dispose(bool disposing) {
  if (disposing) {
    components?.Dispose();
    font?.Dispose();
  }
  base.Dispose(disposing);
  GC.SuppressFinalize(this);
}
```

klightspeed:
I think the base class (which implements the parameterless `Dispose()` method would already do `GC.SuppressFinalize(this)`

phroggster:
Oy. Now you've got me mixing up my dispose patterns. 😃

phroggster:
I've pushed a corrected version of this one, again, just for my sanity or if we want a new release before other options get in.

PR #832, opened 2017-02-15
Release 6 downloadform font handle (Branch_Release_6.0)

phroggster:
Cleanup a font handle leak in DownloadManagerForm

    (cherry picked from commit 4dc6e0a4f8643aaee0a990429c523487a272c17e)

### DIFF
--- a/EDDiscovery/Forms/DownloadManagerForm.Designer.cs
+++ b/EDDiscovery/Forms/DownloadManagerForm.Designer.cs
@@ -17,6 +17,7 @@
             {
                 components.Dispose();
             }
+            font?.Dispose();
             base.Dispose(disposing);
         }
 


EDDiscovery/Forms/DownloadManagerForm.Designer.cs
```diff
@@ -17,6 +17,7 @@ protected override void Dispose(bool disposing)
             {
                 components.Dispose();
             }
+            font?.Dispose();
```
klightspeed: Should `Dispose(bool disposing)` be moved out of the designer code, so the designer doesn't try to rewrite it?

Also, the font should only be disposed if `disposing` is true, as if `disposing` is false the `Dispose` method is being called by the finalizer, and it's not safe to directly dispose of the font (which would probably itself be being finalized).

phroggster: I've never seen the designer re-write anything outside of the commented region that tells you not to modify that which is contained inside the commented region. Dispose is safe in that regard, as designer intentionally uses components to avoid modifying dispose after is created from boilerplate.

As for the second paragraph... You are correct. I was playing around with this in a code base that was modified to dispose of this form via a using block, and I didn't even think about the repercussions of not also including that change.

PR #839, opened 2017-02-18
Add Color-ful object extensions

phroggster:
Note that these are slightly more expensive than the unchecked `byte` under/overflow-ridden methods that `ButtonExt`, etc., were previously providing. But, these methods are extensively tested, well documented, and openly accessible.

- `Color.Average(Color other, float ratio = 0.5f);` where `0.0f <= ratio <= 1.0f`. Averages two Color objects together, complete with alpha, with ratio determining the ratio of the original color to the other Color. Throws `ArgumentOutOfRangeException()` if ratio is `float.NaN`.
- `Color.Multiply(float amount = 1.0f);` where `0.0f <= amount <= ∞`. Multiply a color. That is, brighten or darken all three channels by the same amount, without modification to the alpha value. Returns the original color if amount is `float.NaN` or an otherwise useless value.
- `Color.IsEqual(Color other);` Returns `true` if both colors are 100% transparent, or are otherwise identical. Explicitly ignores any useless name comparisons, unlike the default color equality operator.
- `Color.IsFullyTransparent();` Returns `true` if the color has 0 alpha, `false` otherwise. Hopefully, this will put an end to the `if (someColor == Color.Transparent)` checks that are extremely naughty in some circumstances.
- `Color.IsSemiTransparent();` Returns `true` if alpha is less than 255, `false` otherwise.

Due to the cost increases with Average() and Multiply(), it's suggested to cache these values locally and update them when an input value changes, but you should have already been doing that because they were already "expensive" for an OnPaint() method.


robbyxp1:
Good idea.. the positioning of the functions was always a bodge i meant to solve at some point.

finwen:
Conflict that needs to be fixed before merge

PR #856, opened 2017-02-26
Move DB init out of main form init

klightspeed:
* Start splash first
* Initialize database before main form creation
* Close splash when main form is shown


phroggster:
I feel like this should be going into master sooner than this branch will likely get there.

PR #871, opened 2017-03-02
Figured out where all of my vRAM was going...

phroggster:
OpenTK apparently expects to be disposed of, as it lacks context inside of the finalizer. Also, don't use the 3D map with a debug build of OpenTK or GLControl if you value your sanity.

### DIFF
--- a/EDDiscovery/Program.cs
+++ b/EDDiscovery/Program.cs
@@ -35,7 +35,7 @@ namespace EDDiscovery
         [STAThread]
         static void Main()
         {
-            OpenTK.Toolkit.Init(new OpenTK.ToolkitOptions { EnableHighResolution = false });
+            var tk = OpenTK.Toolkit.Init(new OpenTK.ToolkitOptions { EnableHighResolution = false });
             Application.EnableVisualStyles();
             Application.SetCompatibleTextRenderingDefault(false);
 
@@ -57,6 +57,7 @@ namespace EDDiscovery
                  * TODO: show a dialog and/or bring the current instance's window to the foreground.
                  */
             }
+            tk.Dispose();
         }
 
         static void Run()


klightspeed:
So is the process kept alive during a debugging session, or is Windows vRAM memory management not up to scratch?

phroggster:
Here's `OpenTK.~Toolkit` if it helps to figure out the problem, but yeah, my guess is Windows:
```CS
#if DEBUG
/// <summary>
/// Finalizes this instance.
/// </summary>
~Toolkit()
{
    Debug.Print("[Warning] {0} leaked, did you forget to call Dispose()?");
    // We may not Dispose() the toolkit from the finalizer thread,
    // as that will crash on many operating systems.
}
#endif
```
We *should* probably be `using()` it for a slight bit more safety during debugging or for crashes, force closes, process kills, etc... I'll adjust the PR.

phroggster:
To be clear, I don't notice the vRAM leak in an unmolested master,  just while screwing around with patently absurd ideas in the map. Not sure why Windows doesn't free the vRAM appropriately when the application ends, but I'm going to say that it's probably the Nvidia drivers trying to hold onto it. At least with an explicit `Dispose()` (or even better, an `using()`), we don't have to worry about it.

klightspeed:
Crashes, force-closes, process kills, cutting power, etc. are some of the times when the implied `finally` won't be run, because the process needs to be alive to run the `finally`.  Also, it won't be run after the `WM_ENDSESSION` message has been processed during shutdown.

phroggster:
Wait, what do you mean the (implied) `finally` doesn't run if the power gets cut in the middle? I thought that it was guaranteed to run! ;-) We should probably roll an UEFI image to handle these scenarios. As far as cutting power, that's up to the hardware team to solve.

PR #895, opened 2017-03-07
Master moreicons

robbyxp1:
Many more icons, all done.  4 events which hardly ever appear not done on purpose


robbyxp1:
Many more icons. Phew!

PR #920, opened 2017-03-22
Use an ApplicationContext; close splash after startup

klightspeed:


klightspeed:
I'll rebase this onto master.

PR #934, opened 2017-03-30
Merge branch 'master' into master-configchanged-events

klightspeed:


phroggster:
Feel free to commit directly to the branch. Still trying to figure out the best path forward to convert the config properties (default value, eventhandler, get/set) down to one or two lines of code each. Kind of stuck for a bit on this, unless I also gut OnNotifyPropertyChanged (and all hope of direct UI property binding, which is what started me down this path in the first place).

Issue #933, opened 2017-03-30
Unhandled UI Exception


BlueDrache:
![objection](https://cloud.githubusercontent.com/assets/7607787/24508549/5c208b94-1529-11e7-8bee-63de5d1782bd.png)

BlueDrache:
[Trace_20170330090753.0 - Copy.log.txt](https://github.com/EDDiscovery/EDDiscovery/files/882591/Trace_20170330090753.0.-.Copy.log.txt)

klightspeed:
@finwen is EDDiscovery.pdb not included in the install?  The lack of line numbers in the stacktrace suggests that either it isn't included or doesn't match the built executable.

klightspeed:
@BlueDrache Does the error still occur when you use the [portable version](https://github.com/EDDiscovery/EDDiscovery/releases/download/Release_6.1.3/EDDiscovery.Portable.zip) and copy everything from `%LocalAppData%\EDDiscovery` to the `Data` directory in whichever directory you extract the portable version?

Also, was this a once-off error, or does it occur every time you select the entry you selected when this error occurred?

BlueDrache:
Sorry it took me so long.  I work nights so I run on, essentially, Australian time.  Anyway.  I was able to replicate a glitch but it didn't quite overload and throw an exception.

I'm guessing it's caused by the length of string or amount of data in the journal field "Information" for collecting the sold systems that get dumped in on a mass-sale.  I came back and sold max and it's wigging out when I attempt to display.  Slows system to a crawl when passing over the elements in the viewer.

Will export and post CSV of systems sold journal in date range of issue for replication.

i7-2600 3.4GHz 16GB RAM GTX950 2GB 1.5T HDD  Win 7 Home Pro


[AllExplorationSold.csv.txt](https://github.com/EDDiscovery/EDDiscovery/files/885519/AllExplorationSold.csv.txt)

Rename to .csv and open in favourite spreadsheet.

robbyxp1:
Thanks to another user.. found it. See PR 940

PR #940, opened 2017-04-03
Using CellEnter would cause a UI crash if cursor keys held down

robbyxp1:
Due to cell enter being called by the underlying system when the grid
row was filled, and the tag not being set up yet.

So, don't use cell enter (it was a clodge anyway to get the cursor keys
working).  Instead use key down/up and use that to decide when to
update.


robbyxp1:
See issue #933

PR #942, opened 2017-04-04
Delay committing system note while editing

### LINKED ISSUE #941: [6.2.0Test1] Editing note field is unusably slow
Commit 0ceb8bbe1ac33530566dce18801dae39e6561c18 refactored Note Editing. This change causes every keypress to trigger a database update, which makes entering/editing notes unusably slow.

Removing the call to StoreSystemNote() from richTextBoxNote_TextChanged() event fixes this (although may result in other unwanted behaviour).

```
diff --git a/EDDiscovery/TravelHistoryControl.cs b/EDDiscovery/TravelHistoryControl.cs
index 75082fd..f7526c2 100644
--- a/EDDiscovery/TravelHistoryControl.cs
+++ b/EDDiscovery/TravelHistoryControl.cs
@@ -631,7 +631,7 @@ private void richTextBoxNote_Leave(object sender, EventArgs e)

         private void richTextBoxNote_TextChanged(object sender, EventArgs e)
         {
-            StoreSystemNote(false);
+            //StoreSystemNote(false);
         }
```

klightspeed:
Should fix #941

### DIFF
--- a/EDDiscovery/EDDiscoveryForm.cs
+++ b/EDDiscovery/EDDiscoveryForm.cs
@@ -97,8 +97,7 @@ namespace EDDiscovery
         private int _formTop;
         private int _formLeft;
 
-        private System.Windows.Forms.Timer _delayedNoteStoreTimer;
-        private HistoryEntry _delayedNoteStoreHistoryEntry;
+        private HistoryEntry _uncommittedNoteHistoryEntry;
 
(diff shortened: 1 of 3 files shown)


mwerle:
Is there a reason we need to commit the note while still editing it?  Wouldn't it be better to just commit when done?

PR #944, opened 2017-04-06
Fix json error on saving null user paths

klightspeed:


klightspeed:
Identical fix pushed to master

PR #968, opened 2017-04-16
Update EDSM sync for new rate and time limits

klightspeed:
This updates the log fetch from EDSM to paginate the log fetch requests, as required by the updated API.


finwen:
Shall we pull this into 7.0.x now too?

Issue #983, opened 2017-04-18
Lag when entering Notes


adermon:
There is a delay of about 1 to 8 seconds when entering notes. To be precise the character entry delay between pressing a key on the keyboard and it appearing in the Notes window has variable delay that makes it difficult to use. This was occurring in 6.1.3 at random times but seems worse now in 7.0.

mwerle:
Possible work-around: don't open the 3dMap. (Even opening it and closing it probably causes this issue).

finwen:
should be solved in 7.0.3

PR #988, opened 2017-04-18
Fix slow note editing when 3dmap is open

mwerle:
Move updating the Map Note from every keypress to only on exiting the control as otherwise it is unusably slow.


robbyxp1:
Hang on.. did not Bravada fix this already?

klightspeed:
My fix just prevented saving the note to the database and sped up the search of the journal entry in the history list.  It looks like this fix prevents updating the note in the 3D map until it would be saved to the database and/or sent to EDSM.

mwerle:
Looks like another user has reported it - issue #983

PR #996, opened 2017-04-19
Added Materials filter to engineering panel

iainross:
Also fixed a couple of recipe typos and got rid of the excessively large switch statement mapping UI names to journal names for encoded mats.


iainross:
bugger, changed at the same time as @finwen 

Sorting conflict....

iainross:
Should be OK now, I don't think I've broken what @finwen  was trying to do...

Issue #1016, opened 2017-04-24
System Note is no longer sent to EDSM


mwerle:
Since the recent changes to improve system note performance, it is no longer sent to EDSM.

finwen:
Is this solved now?

mwerle:
Yes, just spent the evening testing and it's fixed (in master, will need to push to release branch).

PR #1017, opened 2017-04-24
[bug #1016] Fix for System Note no longer being sent to EDSM.

mwerle:
Due to the way the code was refactored and now works, the logic to send the note to EDSM was never triggered.

This (hopefully) fixes the notes once and for all (until we rework them to account for recent enhancement requests anyway...)


mwerle:
Merged too soon - realised I made a mistake while lying awake in bed this morning.

PR #1020, opened 2017-04-25
Oops - don't overwrite uncommitted System Notes!

mwerle:


EDDiscovery/EDDiscoveryForm.cs
```diff
@@ -936,7 +936,10 @@ public void StoreSystemNote(HistoryEntry he, string txt, bool send = false)
                 // Also notify various "quick" listeners
                 if (he.UpdateSystemNote(txt, send))
                 {
-                    _uncommittedNoteHistoryEntry = he;
+                    if (_uncommittedNoteHistoryEntry == null)
```
klightspeed: Perhaps only set `_uncommittedNoteHistoryEntry` if `send == false`, then break the storage and sending to its own function, and call that with the `he` when `send == true`?

mwerle: Only need to set it once since it's a reference and it will be automatically sent in the next section if send==true. I wanted to avoid duplicating code since we still need the check for an uncommitted note (although not 100% sure whether that can actually happen now).


Overall I think the whole note-saving probably needs a bit of a refactor, especially if we implement the per-jump notes etc. For now I wanted to make the least possible changes to avoid introducing new bugs (and failing...)

klightspeed: I guess all of the conditions that would cause a new note to be sent would first cause the text box to be unfocussed and therefore cause the old note to be sent.

Issue #1025, opened 2017-04-26
Unhandled exception on start


Artemis387:
For the past few months I haven't been able to launch EDDiscovery, even when updated to the latest version.

The log shows this about the issue:

[2017-04-26 17:54:16Z] Running logfile age check
[2017-04-26 17:54:16Z] Native library pre-loader is trying to load native SQLite library "C:\Program Files\EDDiscovery\EDDiscovery\x64\SQLite.Interop.dll"...
[2017-04-26 17:54:16Z] SQLite version 3.15.2
[2017-04-26 17:54:16Z] SQLite error (1): no such table: register
[2017-04-26 17:54:16Z] 
==== UNHANDLED EXCEPTION ====
System.Data.SQLite.SQLiteException (0x80004005): SQL logic error or missing database
no such table: register
   at System.Data.SQLite.SQLite3.Prepare(SQLiteConnection cnn, String strSql, SQLiteStatement previous, UInt32 timeoutMS, String& strRemain)
   at System.Data.SQLite.SQLiteCommand.BuildNextCommand()
   at System.Data.SQLite.SQLiteDataReader.NextResult()
   at System.Data.SQLite.SQLiteDataReader..ctor(SQLiteCommand cmd, CommandBehavior behave)
   at System.Data.SQLite.SQLiteCommand.ExecuteReader(CommandBehavior behavior)
   at EDDiscovery.DB.SQLiteDataReaderED`1..ctor(DbCommand cmd, CommandBehavior behaviour, SQLiteTransactionED`1 txn, SQLiteTxnLockED`1 txnlock)
   at EDDiscovery.DB.SQLiteCommandED`1.ExecuteDbDataReader(CommandBehavior behavior)
   at EDDiscovery.DB.SQLiteConnectionED`1.GetRegister(Dictionary`2 regs)
   at EDDiscovery.DB.SQLiteConnectionUser.EarlyGetRegister()
   at EDDiscovery.EDDiscoveryController.InitializeConfig(Boolean noreposition)
   at EDDiscovery.EDDiscoveryController.Initialize(Boolean noreposition)
   at EDDiscovery.Forms.SplashForm.Init()
   at EDDiscovery.Program.Run()
   at EDDiscovery.Program.Main()
==== cut ====

klightspeed:
Please look in `%LOCALAPPDATA%\EDDiscovery`.  The error suggests that `EDDUser.sqlite` exists but has no data.

[...]

PR #1026, opened 2017-04-26
Catch errors reading register from EDDUser.sqlite

klightspeed:
Don't error out when EDDUser.sqlite exists but has no data.  This should prevent a repeat of #1025.


klightspeed:
Could we please cherry-pick this commit into 7.0 to fix #1052

PR #1032, opened 2017-05-03
Removed netlogdir from codebase.. #if out CommaderPaths.json

robbyxp1:
Removed old loading of commanders from register
Removed netlogdir - written to db as an empty field. Not removing from
db.
Tested re-import of journals and old netlogs successfully.  Admin option
is still there.

### DIFF
--- a/EDDiscovery/EliteDangerous/EDCommander.cs
+++ b/EDDiscovery/EliteDangerous/EDCommander.cs
@@ -219,11 +219,11 @@ namespace EDDiscovery
 
         public static EDCommander Create(EDCommander other )
         {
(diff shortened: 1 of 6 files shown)


EDDiscovery/EliteDangerous/EDCommander.cs
```diff
@@ -316,6 +316,8 @@ public static void Update(List<EDCommander> cmdrlist, bool reload)
                 }
             }
 
+            #if false       //TBD what is this code for?  turned off due to it not being understood
```
klightspeed: CommanderPaths.json was created in response to some people sharing their user DB between different computers and having different paths to their journals on those computers.

klightspeed:
So someone upgrading from before 5.0 will need to upgrade first to 7.0.x or earlier and migrate their database in order to bring over any commanders from that earlier version?  Might as well rip out SQLiteConnectionOld in that case.

PR #1037, opened 2017-05-06
Fix tests broken by c9c3654

klightspeed:
Commit c9c3654 changed the names of some filters, which caused some tests to fail.  Fix these tests to check the new names.


robbyxp1:
ooooppss always forget about those.

PR #1077, opened 2017-06-07
Migrate Controls namespace to EDDiscovery.ExtendedControls

phroggster:
Add a DoubleBufferedDataGridView class that acts mostly identical to a DataGridView.
Migrate Controls namespace to EDDiscovery.ExtendedControls.
Retire DataGridViewExtensions.
Lastly, organize the compile sections in EDDiscovery.csproj.


robbyxp1:
So, I'm not keen on changing the namespace.  The controls are just additional controls for any program, they just happen to be in EDDiscovery.. Why should they be in that namespace.

Also, can we hold this off until we have 8.X on a stable branch..

finwen:
Will wait for 8.1.x should get out 8.0.x soon first.


2017-06-08 11:26 GMT+02:00 robbyxp1 <notifications@github.com>:

robbyxp1:
I don't like the namespace renaming.   These are generic controls, not EDDiscovery controls.   I use them in other programs, with them copied directly between the programs so to avoid sync issues.

Please, can we do the push without the namespace change.

phroggster:
Aye aye, wasn't aware that they were used externally. That makes me wonder if spinning them off into a separate control library would be the right way to go.

It also begs the question: for the Controls that are explicitly dependant on EDDiscovery application logic (like the auto-complete datagridview systems column, or the HistoryEntry/HistoryList-aware least-cost auto-numbering DGV that I'm working on, and perhaps a couple of others), would such Controls be better off moved out of their current namespace and into some other namespace inside of the EDDiscovery root namespace?

robbyxp1:
I agree they could be segregated more from the EDD code.  The reason i've not DLL them or anything like that is because they seem to be always in development !  So it was just easier to keep them in the project.

But we should make them not dependent on stuff outside of them, as they are supposed to be just better windows form controls

PR #1085, opened 2017-06-15
ExtendedControls visits the library (gets severed off into it's own control library)

phroggster:
There was quite a bit of manipulation needed to separate this namespace cleanly, so several interfaces have seen tweaks. Notably:

* All events now have a `protected virtual void OnEvent(EventArgs)` dispatcher.
* All events previously named `OnEvent` had their library cards revoked, and were ejected from the library party.
* `#region` somehow got wildly abused during the `OnEvent` ejection. Hopefully it has already repressed the memories.
* TabStrip event and delegate names were clarified.
* `protected` event handlers (!!!) became `private` event handlers that raised new events (where needed).
* Numerous `Args` became `EventArgs`.
* Lastly, plugged many leaks, notably `OnPaint() GraphicsPath` items, but there's quite a few others. Yay!

AssemblyInfo was left pretty bare.


robbyxp1:
Good, just hold this for a while.  I don't think the master is clear yet of 8.0.x.  And i want to incorporate my next massive push first before we change this low layer

phroggster:
Take note of the `OnPaint()` cleanups, particularly the fact that the `RectCutCorners()` return needs disposed of, and there was one somewhere that created it's own graphics context that wasn't cleaned up. Those changes need to get integrated ASAP, but it'll be a few days before I can get a more targeted, less obtrusive, PR sent in for them.

Did you play around with the control library? Does that approach work at all for the other applications that are `using Extended controls;` ? It seems like it's the right long-term approach for these Lego blocks, at least to me.

phroggster:
Merged into EDDiscovery/[extendedcontrols_isolation](/EDDiscovery/EDDiscovery/tree/extendedcontrols_isolation) branch via #1090. The changes noted in my previous comment regarding `OnPaint()` cleanups has been split off into #1091.

PR #1092, opened 2017-06-23
Make main form resizes / drags / hittests smarter

phroggster:
* You can now resize appropriately by dragging additional window edges/corners.
  * Drag client (or the giant info label) to top of screen => `Maximize`.
  * Drag client (or the giant info label) from maximized top of screen, in downward direction => `Normal`.
  * BottomRight drag rect is a little larger.
* `ApplyTheme()` ignores `DrawnPanelNoTheme` **parents**.
* MainForm info panel is now a `DrawnPanelNoTheme` (yay, font is legible!)
* About dialog shows concise version number ("eddisco v8.1.0.0", TYVM).
* Removed `buttonReloadActions`, now using `paneleddiscovery_Click`.
* Removed `label_version`, now using `VersionNumMenuItem`.
* Clicking the info panel with update msg shows update form.
* Note the revised NCHITTEST code that can survive without `unsafe()` and won't throw IntPtr overflows, and works even in multi-monitor, negative X/Y-coordinate-land:
```CS
short x = (short)m.LParam.ToInt32();
short y = (short)(m.LParam.ToInt32() >> 16);
Point p = PointToClient(new Point(x, y));
```

Potential Pitfalls:
* Mono users (@klightspeed, you've got a Mac, right?) please review & test, but it **should** be smoother.
  * Everything here is now using C# native mouse handling, except for the hittests.
* The `WndProc` hittest can't seem to pull the top-left corner in the menu margin. Not sure how to address this one.

I really wish that we could use aerosnap with a true no border frame, but these changes should make it behave a little more sanely, and at least satisfy the maximize/de-maximize aspect.

### DIFF
--- a/EDDiscovery/EDDConfig.cs
+++ b/EDDiscovery/EDDConfig.cs
@@ -364,6 +364,12 @@ namespace EDDiscovery
 
         public class OptionsClass
         {
+            /// <summary>The version number of the assembly, formatted as: <c>"8.1.2.3"</c></summary>
+            /// <seealso cref="VersionDisplayString"/>
+            public string VersionNumberShort { get; } = Assembly.GetExecutingAssembly().FullName.Split(',')[1].Split('=')[1];
+            /// <summary>The version number of the assembly, with (some) processed option details:
+            /// <c>"Version 8.1.2.3 (using /some/super/long/appfolder/path) (EDSM No server) (no BETA detect)"</c></summary>
+            /// <seealso cref="VersionNumberShort"/>
             public string VersionDisplayString { get; private set; }
             public string AppFolder { get; private set; }
             public string AppDataDirectory { get; private set; }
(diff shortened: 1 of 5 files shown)


phroggster:
Erm... I forgot to account for virtual monitors located above the primary (resulting in negative Y positions) in a couple of places. Let me get that fixed before this goes in.

Negative X positions work fine right now, but anyone with a vertical monitor stack would probably want to internet-murder me for this.

klightspeed:
Looks like you would use `Screen scn = System.Windows.Forms.Screen.FromPoint(ptscn)` to get the screen, then use `scn.Bounds.Top` and compare Y with that.

phroggster:
Yeah, `HT.CAPTION` was previously responsible for movement, but it actually interferes with the aerosnap movement. `OnMouseMove` / `infoPanel_MouseMove` (both of which are essentially identical) can have more control over the proceedings than what we had merely adjusting hittest results. Additionally, the lack of `base.WndProc(m)` calls was causing certain mouse clicks to get swallowed and never passed into `OnMouseDown()`-land. This approach avoids that, to say the least.

By the way... The new hotness of the day is:
```CS
if (m.Msg == WM.NCHITTEST) {
  Point p = PointToClient(new Point(m.LParam.ToInt32()));
}
```

phroggster:
The 62e48b0 revision fixes the problems that were evident in the first go-around. Please, try this patch out and see how she works on some various .NET implementations and windowing environments.

phroggster:
Closing. Will provide a base class implementation that covers all of our bases within a couple of days.

PR #1143, opened 2017-07-31
Enhance Portable Zip Build and Cleanup Main csproj

phroggster:
Changes to lines 1-1267 are merely cleanup, with no functionality differences.

* Delete EDDiscovery.portable.zip during a clean
* Use wildcards to pull in all .dll and .pdb files from build
* Use `@(SQLiteInteropFiles)` to reduce hard-coded paths
* Use `$(OutDir)` instead of `$(OutputPath)`, which includes the trailing slash
* Zip should now be safe for unix builds (**please test!**)


phroggster:
Problems under xbuild and msbuild on Unix will require re-work. Thanks for checking for me. Will go back to != Unix conditional, and also add a message for the CI logs detailing what is being packaged and from where.

PR #1149, opened 2017-07-31
Enhance Portable Zip Build and Cleanup Main csproj

klightspeed:
Modification of @phroggster's #1148

* Delete EDDiscovery.portable.zip during a clean
* Use wildcards to pull in all .dll and .pdb files from build
* Use @(SQLiteInteropFiles) to reduce hard-coded paths
* Use OutDir instead of OutputPath, which includes the trailing slash
* Re-add SQLite interop

Should help guard against issues like #1136 in the future.

Issue #1136, opened 2017-07-30
Invalid EXE after builiding in Appveyor CI


andreaspada:
Not know why, but by many weeks I can't build anymore my commitment in appveyor as the resulted EXE, although the build was successful, is not valid: It do not have the icon applied, and it do not work at all.

How are you compiling the source tree?

klightspeed:
It looks like the csproj was not updated to include all of the new DLLs in the portable build.

Additionally, there seems to be an error during init when starting from scratch.

andreaspada:
Wow... indeed! Thanks and congratulations for the super fast fix! KUDOS

Enviado do meu iPhone

robbyxp1:
Master is in heavy flux note. expect more problems!

phroggster:
Once #1143 gets verified as Unix-safe and merged, this issue should be addressed in a more permanent and future-friendly form.

PR #1150, opened 2017-08-01
Enhance Portable Zip Build

klightspeed:
Modification of #1149, with input from review on that PR, minus the csproj cleanup commit.

* Delete EDDiscovery.portable.zip during a clean
* Use wildcards to pull in all .dll and .pdb files from build
* Use @(SQLiteInteropFiles) to reduce hard-coded paths
* Use OutDir instead of OutputPath, which includes the trailing slash
* Re-add SQLite interop

This should get the portable zip working again, and prevent further issues (such as #1136) due to adding or removing DLLs.

### DIFF
--- a/EDDiscovery/App.Portable.config
+++ b/EDDiscovery/App.Portable.config
@@ -33,12 +33,6 @@
       </providers>
     </roleManager>
   </system.web>
-  <system.data>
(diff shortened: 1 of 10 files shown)


mwerle:
Works For Me now.

PR #1153, opened 2017-08-01
A few more assembly name / version corrections

phroggster:
Also, add a proper UserAgent for GitHub requests (was TestApp). See the [API docs](https://developer.github.com/v3/#user-agent-required).
Fix first chance exception handler missing out on exceptions
Reduce number of times that we process the version number/assembly name throughout, and just keep a few static strings.


robbyxp1:
Good work

PR #1159, opened 2017-08-02
Fix toggled case causing mono build failure

klightspeed:
Fixes 202f806

### DIFF
--- a/TestExtendedControls/TestExtendedControls.csproj
+++ b/TestExtendedControls/TestExtendedControls.csproj
@@ -117,7 +117,7 @@
     </Compile>
   </ItemGroup>
   <ItemGroup>
-    <None Include="app.config">
+    <None Include="App.config">
       <SubType>Designer</SubType>
     </None>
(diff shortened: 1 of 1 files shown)


phroggster:
It seems odd that xbuild treats app.config properly in all of the other assemblies, but wants to be case-sensitive for this one... Wasn't sure whether this would appease the build $deity or merely make it angrier. Apparently the ~~human~~ sacrifice worked.

klightspeed:
Don't know why Visual Studio changed it in the first place, unless it was accidentally removed and re-added by typing the name.

phroggster:
I'm guessing that it has to do with the packages being updated for the referenced projects. Update NuGet packages for, say, BaseUtils, and it trickles up to everything that references BaseUtils. In Visual Studio, it's adamant about having the designer subtype, so it adds it in if it's not there, and it always treats app.config as lower-case. With the subtype there, it shouldn't change it again, at least, that's the theory. ;-)

PR #1160, opened 2017-08-02
Fix CS0114 on ChangeTravelGrid

klightspeed:
Change `virtual` to `override` on overridden methods.


robbyxp1:
I'll do it on my branch.. i have many changes around this

PR #1162, opened 2017-08-03
Cleanup more duplicated resources, Route/Exp icon

phroggster:
Cleanup a couple more duplicated resources
Add a map icon to Routes/Expeditions tab; it was looking lonely


robbyxp1:
Embedded icon resources! No!  I've been removing them because they are only visible to one bit.
may break out.

PR #1170, opened 2017-08-04
Mainform dblclick in non-client toggles maximize

phroggster:
Some clever Timer usage, and a cached MouseEventArgs, allows us to toggle maximized state when double-clicking on a caption-esque control.

NOTE: this changes the behavior of of the caption controls to only handle mousedown when it's the left mouse button and the form is drawn without the window frame. This may not neccessarily be desired, but is, IMHO, the correct thing to do.

NCLBUTTONDBLCLK is not used, but was added during testing and is probably a good thing to hold onto.

### DIFF
--- a/BaseUtils/Misc/Win32Constants.cs
+++ b/BaseUtils/Misc/Win32Constants.cs
@@ -187,6 +187,14 @@ namespace BaseUtils.Win32Constants
         public const int NCLBUTTONUP = 0x00A2;
 
         /// <summary>
+        /// The WM_NCLBUTTONDBLCLK message is posted when the user double-clicks the left mouse button while the cursor
+        /// is within the nonclient area of a window. This message is posted to the window that contains the cursor. If
+        /// a window has captured the mouse, this message is not posted.
+        /// </summary>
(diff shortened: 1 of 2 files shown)


phroggster:
Maybe now I'll quit double-clicking for 30 seconds to restore the window from maximized when it's themed. 🥇

robbyxp1:
Can we pend this for a bit.. need to experiment with it first

phroggster:
Yeah, take your time. Frankly, I'd prefer the non-client area expansion with custom drawing to cover up the title bar route, but that's damn hard to do correctly (it does allow for Aerosnap, something else that I miss). The other option is to handle window movement ourselves, which I think I proposed at some point.

The main thing is that I'd like for our non-bordered form to behave exactly as if it was bordered: Aerosnap, double-click maximize/restore, etc., minus the ugly border and caption colour. Anything closer to that is a good thing and should be pursued.

robbyxp1:
Agreed. Just want to be a bit more cautious what we pull in now. After 8.2, we can open it up for more stuff.

phroggster:
I've started pushing my luck a little more trying out the [NonClient route](https://github.com/phroggster/EDDiscovery/tree/nonclient_totalboss). It's near to completion, and I think it's the correct way to move forward, even if it's using the TestExtendedControls approach.

PR #1189, opened 2017-08-11
Add a FatalExceptionForm : Form

phroggster:
Displays an exception similar to the debug unhandled exception dialog.
Provides a one-click copy details to clipboard and open github/issues button.

Migrate EDDApplicationContext init routine to use this dialog, which...
Fixes the old dialog not showing line numbers or InnerExceptions.

Several other places could probably use this dialog, as well as a non-fatal variant. Will have to see where else to recycle this...

Screenshots of some random exception that I threw in for testing:

### DIFF
--- a/EDDiscovery/EDDApplicationContext.cs
+++ b/EDDiscovery/EDDApplicationContext.cs
@@ -14,6 +14,7 @@
  * EDDiscovery is not affiliated with Frontier Developments plc.
  */
 using EDDiscovery.Forms;
+using ExtendedControls;
 using System;
 using System.Diagnostics;
 using System.IO;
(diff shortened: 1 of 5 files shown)


klightspeed:
Looks eerily like the old Internet Explorer javascript error dialog.

phroggster:
Well, yeah, that was the inspiration. Not Internet Explorer, but the near-identical exception dialog that MSVCRTd has been using for quite a number of years. I'm kind of hoping that the ~~ugliness~~ sheer utility helps us to handle more exceptions closer to the hot plate where we might actually be able to recover or workaround.

PR #1198, opened 2017-08-14
Safemode Implemented

robbyxp1:
DB move implemented.. changed method sorry to a simpler one of placing an options file with the paths in the appfolder.  Tested moving it around.  Cleaned up options file as I found it was not reading the options in the right order and to add in the reading of a fixed option file for db movement.


EDDiscovery/EDDOptions.cs
```diff
@@ -113,110 +107,43 @@ private void ProcessConfigVariables()
...
 
-        private void ProcessOptionsFile()
+        private void ProcessOptionsFileOption()     // command line -optionsfile
         {
             string optionsFileName = "options.txt";
-            bool useAppDataOptionsFile = false;
 
-            ProcessCommandLineOptions((optname, optval) =>
-            {
+            ProcessCommandLineOptions((optname, optval) =>              //FIRST pass thru command line options looking
```
klightspeed: This removes the implicit `options.txt` that has been present for the past few versions.

PR #1202, opened 2017-08-14
Cherry-pick some stuff over from my nonclient_totalboss branch

phroggster:
Cherry-pick some immediately useful stuff over from my [non-client / DWM rendering branch](https://github.com/phroggster/EDDiscovery/tree/nonclient_totalboss).

* Excessively more win32 constants.
* Add a chooser form for TestExtendedControls (image below)
* DrawnPanel:
  * Add Max/Restore buttons (no, they are not updated automatically)
  * Specify default values and event to improve designer experience.
  * Invalidate when ImageSelected changes.

ChooseTestsForm:
untitled

Elite-themed 24x24px buttons:
untitled

Visual Studio preview of restore button @ 32x32px (others 24x24px):
untitled

### DIFF
--- a/BaseUtils/Misc/Win32Constants.cs
+++ b/BaseUtils/Misc/Win32Constants.cs
@@ -16,6 +16,87 @@
 
 namespace BaseUtils.Win32Constants
 {
+    #region window Class Styles (CS_)
+
+    /// <summary>
+    /// Window Class Style message constants, as defined in Winuser.h (via Windows.h). Useful for manipulating
+    /// <see cref="System.Windows.Forms.CreateParams.ClassStyle"/>.
+    /// </summary>
+    /// <seealso href="https://msdn.microsoft.com/en-us/library/windows/desktop/ff729176(v=vs.85).aspx"/>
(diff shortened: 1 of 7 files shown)


robbyxp1:
Hi,

Sorry, but don't like the choose test forms, if for the primary reason that i just want it to run as quick as possible, without clicking it, and am happy just to mod program.cs each time to pick the form..

Its a good idea, but it needs to be persistent so it just runs.

Whats the two new icons for?

robbyxp1:
Got it, put in a command line option -testformx etc to go directly to the form you are testing.  if no command line, present the form

phroggster:
The command-line/debug executable parameters is a reasonable approach. I'd just like to *not* have to change the code to run a different form. That, and I'd like to merge my main non-client branch in at some point for testing. Until y'all have had a chance to click and drag and play around with it, it make more sense to leave it and a derived Form in the TestExtendedControls project before moving it over to ExtendControls where it belongs.

The icons are for resize/maximize. They look silly as separate icons such as in the screenshot, but make much more sense when switching the same control back and forth between the two (and maximizing or restoring the window at the same time). I didn't wire them up, but that's easy enough to do so.

PR #1205, opened 2017-08-16
EDDApplicationContext.legability += 10;

phroggster:
`.Extendability += 7;`
  `.Safety += 9;`

### DIFF
--- a/EDDiscovery/EDDApplicationContext.cs
+++ b/EDDiscovery/EDDApplicationContext.cs
@@ -34,34 +34,16 @@ namespace EDDiscovery
     // Class for managing application initialization, and proud owner of SplashScreen and EDDiscoveryForm. Singleton.
     internal class EDDApplicationContext : ApplicationContext
     {
-        public EDDApplicationContext(bool safemode) : base(safemode ? ((Form)new SafeModeForm()) : new SplashForm())
+        public EDDApplicationContext() : base(StartupForm)
         {
-            if (safemode)
+            if (typeof(SafeModeForm).IsAssignableFrom(MainForm?.GetType()))
             {
-                ((SafeModeForm)MainForm).Run += RunAfterSafeMode;
+                ((SafeModeForm)MainForm).Run += ((p, t) => { GoForAutoSequenceStart(new EDDFormLaunchArgs(p, t)); });
             }
             else
             {
-                Timer timer = new Timer { Interval = 250, Enabled = true };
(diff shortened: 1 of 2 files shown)


robbyxp1:
it actually looks more complicated ;-)

phroggster:
You're right; I added a `.Complexity` modifier to the commit message (and fixed a misleading comment referencing the removed static `.Instance` & squashed).

Look at the gains from this:
* Safety:
  * The excessive null coalesce usage isn't needed (at all), but shields from future mistakes.
  * Same for the more complex `typeof(T).IsAssignableFrom(obj?.GetType())`-style comparisons instead of `obj is T` that take into account type conversion.
* Readability:
  * `void RunAfterSafeMode(bool p, bool t){...}` means what, exactly, and how readable will that signature be in 2 years?
    * `void GoForAutoSequenceStart(EDDFormLaunchArgs args = null){...}` lowers the ambiguity regarding options.
    * I certainly could have named it more intuitively, but it is **literally** responsible for asynchronously handing off application and main UI initialization from *whatever* invoked it to the "automatic initialization sequence" (*T-31 seconds*).
  * Eliminated code duplication (seealso `void SwitchContext(Form)`, `new Timer { Enabled = true, Interval = 250 }.Tick += ...;`).
  * Everything does what it says on the tin that it should do.
* `static Form StartupForm{get;}` allows for dynamically changing what, if anything we will display.
  * That's a great padded perch to implement actual IPC downstream instead of or in addition to the global Mutex lock.

robbyxp1:
never said i did not want to merge it long term... Why did you take the PR away.  Just did not want to do it right now

phroggster:
Aye. Re-opening.

I did a trial run of the end-goal using SharmIPC. At least with that (which internally relies on a mutex and uses a memory mapped file, so still not Mac perfect), it was cleaner to integrate it into an IPC management class, and wire that into a single `using` block, with two function calls wrapped into a single (IF NOT `bool EddIpc.IsAnotherInstanceInUse{get;}` AND NOT `bool EddIpc.TryActivateOtherInstance()` AND `MessageBox.Show(...) == ...`)-style statement inside Program.cs.

The original idea behind wiring IPC in to EDDApplicationContext was that it may need to create a non-visible window in order to do it's duties, and it's the best place to handle (even an invisible) appwindow. I don't believe that's the best path any longer.

phroggster:
Bump: rebased onto master; no more merge conflict.

PR #1208, opened 2017-08-18
AppVeyor disliked that...

phroggster:


klightspeed:
That would be because those are C#7.0 features (dotnet 4.7, VS2017), while Appveyor would only have dotnet 4.6.2

finwen:
Could be good to have C# 6.0 compatibility  for a while

PR #1220, opened 2017-08-23
Add Restore and Maximize ImageTypes to DrawnPanel

phroggster:
* ExtendedControls/Controls/DrawnPanel.cs:
  * Set DefaultEvent (what event the designer inserts for dblclk).
  * Set DefaultValue attribs on all properties.
    * Allows for the "Reset" context menu item inside of the Visual Studio Properties panel.
    * Allows for Visual Studio to ignore serializing the value to .Designer.cs if unchanged.
  * Add Maximize and Restore items to ImageType enum.
    * Paint them appropriately, although Restore is ugly when tiny.
  * Adjustments to `public ImageType ImageSelected {get;set;}` will now `Invalidate()` to request a re-paint.
* Everywhere else:
  * Remove the redundant DefaultValue assignments from *.Designer.cs

Preview:
drawnpanels
The left three are 24x24 with `MarginSize` of 8, 6 (the most commonly used), and 4 (the default) respectively, while the right three are 32x32, 36x36, and 48x48 respectively.


phroggster:
Merging, and a reminder that they are not yet wired in on (any) form, they're just there for when we need them.

PR #1233, opened 2017-08-26
Fix a couple Paint bugs

phroggster:
* ComboBoxCustom:
  * Fix a bug that was reflecting for every Paint call.
* TextBoxBorder:
  * Fix a bug that was reflecting for every Paint call.
* ControlHelpers:
  * Avoid multiple `O(N)` Linq flattenings when one `O(1)` can do it.


robbyxp1:
Silly old Robert. thanks

robbyxp1:
See this.. 

            return icontainers?.FirstOrDefault()?.GetValue(c) as System.ComponentModel.IContainer;  // But IT may be null if no containers are on the form

the reason it was written out in long hand before was that when i was developing it, i needed to step thru each statement to check it was working.  If you see it in long hand, thats probably why it was done that way.

PR #1236, opened 2017-08-29
Added popout control to list estimated scan values for bodies in system

zebranky:


EliteDangerous/Properties/Resources.resx
```diff
@@ -673,4 +673,7 @@
   <data name="starcluster" type="System.Resources.ResXFileRef, System.Windows.Forms">
     <value>..\Resources\starcluster.PNG;System.Drawing.Bitmap, System.Drawing, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a</value>
   </data>
+  <data name="estval" type="System.Resources.ResXFileRef, System.Windows.Forms">
+    <value>..\Resources\estval.png;System.Drawing.Bitmap, System.Drawing, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a</value>
```
klightspeed: Just a nitpick, since this is only used by the UI, shouldn't this be in the UI project?

zebranky: It should, IMO, but I was being consistent with the placement of the other popout control icons.

finwen:
One idea.  SHould we also (optionally?) show the type of the scanned object
in the list.


2017-08-29 6:11 GMT+02:00 zebranky <notifications@github.com>:

PR #1242, opened 2017-09-01
Link StartJump to next system

### LINKED ISSUE #1241: Update scan panel on jump start event, not on arrival event
In my understanding, currently the scan panel is only updated (and downloaded from EDSM) when you are already in the system.

It would be really handy if I could already check what I am getting into before actually arriving. A "jump start" event can only very rarely be interrupted, so I assume it should not interfere.

klightspeed:
This should update any panels to display information for the next system.

This should fix #1241


EoD:
I used the artifact from appveyor and it seems to work fine.

PR #1254, opened 2017-09-07
Fix null-reference exception on first startup (or, presumably, whenev…

mwerle:
…er a new Commander name is found in a log file).


mwerle:
Not sure why one branch of the "if (_commander == null)" test used a new local variable, but as that branch never set the "_commander" variable, the line following the "if()" blew up.

I'm guessing it's a remnant of a refactor.

robbyxp1:
good catch!

Issue #1147, opened 2017-07-31
Small improvement in floating panel/windows


andreaspada:
I fly using mouse and keyboard. I like to have a Summary panel on my left side of the screen which gave me just the distance to my target and which shows scan information. I left it transparent and on top.

Sometime, my mouse pointer goes on that windows, which highlight it, which is quite annoying. I understand that we need to properly highlight a panel if we need to move or change it's setup, but I would suggest to let it be unresponsive to mouse hover unless we click on it (or, better, unless we select it by task selector or windows switch hotkeys).

What do you think about?

phroggster:
I agree that something needs to be done about this. However, it's quite complicated, as the mouse capture gets stolen from EliteDangerous if it enters another window, even if it's a transparent window on top of Elite. Not sure that it's even possible to implement something to solve this, but I'll do some research and see.

klightspeed:
As in WM_NCHITTEST returns HT_TRANSPARENT?

robbyxp1:
We need a way of activating the spanel ;-)  If you flying with your mouse, ED should have captured it and the mouse pointer will not be moving around. In a station, you can minimise the panel (dock option) to stop the mouse pointer in the station getting caught by the window.

phroggster:
ED running in full-screen borderless does capture the mouse, and it works perfectly fine. The problem is when a window gets placed over the top of it, then capturing gets to be problematic.

HT_TRANSPARENT and WS_LAYERED / WS_TRANSPARENT might be part of a solution, but it's only a part. Some clever hacks will probably need to be employed to have it behave sensibly.

Again, I do feel that this is a problem, but any solution has a high chance of making a slightly different problem worse.

klightspeed:
This gets a form that is completely transparent to mouse input:

```
    public partial class Form1 : Form
    {
        const int WS_EX_TRANSPARENT = 0x00000020;

        public Form1()
        {
            InitializeComponent();
            this.AllowTransparency = true;
            this.TransparencyKey = this.BackColor;
            this.TopMost = true;
        }

        protected override CreateParams CreateParams
        {
            get
            {
                CreateParams p = base.CreateParams;
                p.ExStyle |= WS_EX_TRANSPARENT;
                return p;
            }
        }
    }
```

Unfortunately setting and resetting this after window creation requires a call to GetWindowLong and SetWindowLong, as System.Windows.Forms provides no way to change extended styles after window creation.

andreaspada:
[...]

PR #1263, opened 2017-09-11
Transparent mode improvements

robbyxp1:
Added four state transparency control to system, fixed some issues in keyboard naming, make DI keys work with c# key names


andreaspada:
Can't wait! ;)

robbyxp1:
#1147 is addressed with this

PR #1267, opened 2017-09-12
Less taskbar overlap when unframed

phroggster:
DraggableForm now handles the `WM_GETMINMAXINFO` message to prevent overlapping the taskbar when maximizing.

TODO:
- [ ] This needs to be tested on an x86 (non-amd64) operating system. `POINTS` (which is what the official `MINMAXINFO` struct uses) is not always the same as `System.Drawing.Point`, but it *might* be ok?
- [X] Verify Mac compatibility.

### DIFF
--- a/BaseUtils/Misc/NativeMethods.cs
+++ b/BaseUtils/Misc/NativeMethods.cs
@@ -245,7 +245,18 @@ namespace BaseUtils.Win32
         public static extern uint AssocQueryString(AssocF flags, AssocStr str,
            string pszAssoc, string pszExtra, [Out] StringBuilder pszOut, ref uint
            pcchOut);
-   }
+
+        // https://msdn.microsoft.com/en-us/library/windows/desktop/ms632605.aspx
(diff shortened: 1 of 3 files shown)


klightspeed:
Mono on macOS does not seem to implement borderless windows.

I'm fairly sure the `System.Drawing.Point` struct is supposed to be the managed representation of the `POINT` native structure.

phroggster:
I agree on both accounts, but cannot verify. The C++ `LONG` (as used by [`POINT`](https://msdn.microsoft.com/en-us/library/dd162805(v=vs.85).aspx)) should always be the same as a C# `int` (as used by [System.Drawing.Point](https://msdn.microsoft.com/en-us/library/system.drawing.point(v=vs.110).aspx)) - 32 bits - but I've seen numerous differing managed definitions. Would feel better if someone could maximize and restore this build in a non-amd64 environment, preferably with multiple and massive monitors.

I'll edit out the Mac test TODO item, as that *should* be ok.

PR #1297, opened 2017-09-27
Maybe Resolve #1292

### LINKED ISSUE #1292: Action File Editor open error UNHANDLED UI EXCEPTION
This sentence is machine translated from Japanese to English. It may be unreadable.

An error occurs when trying to edit the Action File.
This log file is version 8.2.7, but this error also occurs in 8.3.0.
Is there a solution?
I use OS language in Japanese.
[Trace_20170917140925.0.log](https://github.com/EDDiscovery/EDDiscovery/files/1327922/Trace_20170917140925.0.log)

phroggster:
Not sure if this will resolve #1292 appropriately or not.

- [ ] This needs to be tested in a non-Cyrillic script locale prior to merge, and until then this PR is merely a placeholder to get a neutral AppVeyor artifact.

### DIFF
--- a/ActionLanguage/ActionEditing/ActionPackEditorForm.cs
+++ b/ActionLanguage/ActionEditing/ActionPackEditorForm.cs
@@ -57,6 +57,7 @@ namespace ActionLanguage
 
         public ActionPackEditorForm() 
         {
+            groups = new List<Group>();
             InitializeComponent();
         }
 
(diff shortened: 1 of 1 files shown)


klightspeed:
It looks a lot like `OnResize` is being called during layout.  If this is the case, moving the initialization of `groups` to before `InitializeComponent()` (or even moving the `= new List<Group>` to the declaration of the `groups` field) should fix this.

phroggster:
Yeah, that's exactly it. `Init()` gets called after the first resize when a non-Cyrillic locale is in use. I ran into the same issue when I added the notification area icon to the main form.

This *should* address it (that is, the second patch that I force-pushed), but there may be consequences that I'm unable to check for.

PR #1298, opened 2017-09-27
Resolve #1278

### LINKED ISSUE #1278: Unknown Module type
Unknown module type:
Thrusters Customisation Pink:EngineCustomisation_Pink

phroggster:
Engine colour customizations don't have an ID.


phroggster:
Seealso [EDMC-ignored items](https://github.com/Marginal/EDMarketConnector/blob/d885f9feb08669620571e62f70b8ab8248ab24a9/outfitting.py#L286). Diff between the two (now) is that we've also got `cargobaydoor`.

PR #1299, opened 2017-09-27
Fix ActionFileEditForm

qubeGH:
FIX UNHANDLED UI EXCEPTION
System.NullReferenceException: Object reference not set to an instance of an object.


phroggster:
ありがとう、qubeGHさん。

Issue #1302, opened 2017-09-27
Screenshot convert but refuse to appear in application


andreaspada:
I'm using EDDiscovery 8.3.1.0

The scresnshots are correctly taken, converted and organized, however they do not display in the app window and the log display the following error:

`Error in executing image conversion, try another screenshot, check output path settings. (Exception Could not find file 'F:\Games\Elite Dangerous\Tools\EDDiscovery\Screenshot_0000.bmp'.)`

ED put the screenshots in d:\Users\"$myuser"\Pictures\Frontier Developments\Elite Dangerous

I organize them in f:\Games\Elite Dangerous\Pictures\Screenshots\"$mycmdr"\...\...

Any clue?

klightspeed:
In your trace log, is there an associated stack trace?

phroggster:
Also, there appears to be three different paths listed. One of these is likely incorrect:
* `F:\Games\Elite Dangerous\Tools\EDDiscovery\`
  * Where ED is reportedly saving the screenshots and EDD is looking for them.
* `D:\Users\$USERNAME\Pictures\Frontier Developments\Elite Dangerous`
  * Where ED is likely saving the screenshots
* `f:\Games\Elite Dangerous\Pictures\Screenshots$cmdr`
  * Where you desire the screenshots to go.

phroggster:
When you said,
> I organize them in f:\Games\Elite Dangerous\Pictures\Screenshots"$mycmdr"......

Is that the output path that EDD is set to save the converted screenshots to?

andreaspada:
Well, certainly the path in the error message is wrong. I neither imagine why it's here... I never used it.

---

> When you said,
> 
>     I organize them in f:\Games\Elite Dangerous\Pictures\Screenshots"$mycmdr"......
> 
> Is that the output path that EDD is set to save the converted screenshots to?

Yes. And it's working. Here it is my setup:

![capture-eddiscovery_screenshot_config](https://user-images.githubusercontent.com/4296635/30940957-8d74b21e-a3b9-11e7-986b-8654e1d6c4d1.PNG)

In the log, I find this traces:

```
[2017-09-27 19:19:20Z] Trace:    at System.IO.__Error.WinIOError(Int32 errorCode, String maybeFullPath)

[...]

PR #1318, opened 2017-09-29
Call OnNewJournalEntry first

klightspeed:
This should hopefully fix the OnNewJournalEntry delay seen in #1302


robbyxp1:
Okay.. but what could be holding it up so long.  I mean it can run action programs, but they quit after a period of run time.. what do you think is holding it up?

klightspeed:
Part of it could be the image control taking a few seconds to load the image (likely as soon as the image path is set), especially if it's a big bitmap.

robbyxp1:
well, i was worried it could be the action stuff.. there is no reason why it can't be first

PR #1326, opened 2017-09-30
DateTime.Now is not DateTime.UtcNow

phroggster:
No wonder the tracelogging timestamps have been off, despite having the Zulu indicator.


phroggster:
[Here's a good StackOverflow](https://stackoverflow.com/questions/915693/why-does-datetime-now-tostringu-not-work) on the issue that this is solving.

PR #1329, opened 2017-10-01
Prevent splitter error in shopping panel with extremely narrow non-transparent forms

iainross:
And only react to IMaterialCommodityJournalEntry events

### DIFF
--- a/EDDiscovery/UserControls/UserControlShoppingList.cs
+++ b/EDDiscovery/UserControls/UserControlShoppingList.cs
@@ -182,7 +182,7 @@ namespace EDDiscovery.UserControls
                     splitContainer1.Panel1MinSize = displayList.img.Width;
                 }
                 else
-                    splitContainer1.SplitterDistance = splitContainer1.Width - splitContainer1.Panel2MinSize;
+                    splitContainer1.SplitterDistance = Math.Max(0, splitContainer1.Width - splitContainer1.Panel2MinSize);
             }
         }
 
--- a/EDDiscovery/UserControls/UserControlShoppingList.cs
+++ b/EDDiscovery/UserControls/UserControlShoppingList.cs
@@ -106,7 +106,8 @@ namespace EDDiscovery.UserControls
         private void Discoveryform_OnNewEntry(HistoryEntry he, HistoryList hl)
         {
             last_he = he;
-            Display();
+            if (he is IMaterialCommodityJournalEntry)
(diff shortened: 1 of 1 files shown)


EDDiscovery/UserControls/UserControlShoppingList.cs
```diff
@@ -106,7 +106,8 @@ private void Synthesis_OnWantedChange(List<Tuple<MaterialCommoditiesList.Recipe,
         private void Discoveryform_OnNewEntry(HistoryEntry he, HistoryList hl)
         {
             last_he = he;
-            Display();
+            if (he is IMaterialCommodityJournalEntry)
```
klightspeed: This will never be true.  It needs to be `if (he.JournalEntry is IMaterialCommodityJournalEntry)`

iainross: It's working like that and the display's updating when I pick up materials though...

iainross: Because it's being triggered from below by the synthesis and engineering controls raising their OnWantedChange from their Display methods and those're responding to every history entry so even if this was right it wouldn't achieve what @phroggster was looking for when he asked me to stick this in...

iainross: I'll redo it across all three controls in master in a bit.

phroggster: Does it still work if there's no engineering and synthesis controls in use? If so, then the whole `OnNewEntry` handler can be avoided.

The goal, though, is to minimize calling `Display()` for a bunch of stuff that it doesn't care about. Whether that's a single call after a batch of operations, or filtering out non-matcom events before calling it, the cost savings are very likely worth it.

iainross: The eng and synth controls are embedded in the shopping list, so no.

What I've done on my branch (and not finished testing yet) is to keep the OnNewEntry handler in shopping list but stop the embedded controls also propagating that up.

I think that "Every UC handles history as it sees fit and embedded controls only pass on user generated events" is a better pattern than "UCs with embedded child UCs rely on the children to pass history events up the chain".

The former would result in every MatCom event being sent to the shopping list by both eng and synth controls plus it muddies the water if we change things in future if some UCs don't see history events but others do.

PR #1330, opened 2017-10-01
Prevent splitter error in shopping panel with extremely narrow non-transparent forms

iainross:
And only react to IMaterialCommodityJournalEntry events


EDDiscovery/UserControls/UserControlShoppingList.cs
```diff
@@ -106,7 +106,8 @@ private void Synthesis_OnWantedChange(List<Tuple<MaterialCommoditiesList.Recipe,
         private void Discoveryform_OnNewEntry(HistoryEntry he, HistoryList hl)
         {
             last_he = he;
-            Display();
+            if (he is IMaterialCommodityJournalEntry)
```
phroggster: I think it's `if (he.journalEntry is IMaterialCommodityJournalEntry)`

PR #1332, opened 2017-10-02
Rewrite the TraceLog implementation (master)

phroggster:
* "Previous line duplicated" messages are smarter, simpler.
* Console and Trace loggers now use an identical TextWriter
  * No more races between the two logging mechanisms.
* Exception newlines are now `Environment.NewLine` instead of '\n'.
* Fully documented the class, not that it is ever used outside of Init.
  * Seriously, this class has only two public methods.
* C:\*.log won't be deleted when things go awry (yes, seriously, this *was* a possibility with the old implementation).
* Maximum log directory size (formerly `MaxLogDirSizeMB`, now `MaxLogDirSizeMiB`) reduced from 1 GiB to 100 MiB.

### DIFF
--- a/BaseUtils/Misc/TraceLog.cs
+++ b/BaseUtils/Misc/TraceLog.cs
@@ -14,142 +14,371 @@
  * EDDiscovery is not affiliated with Frontier Developments plc.
  */
 using System;
-using System.Collections.Generic;
 using System.Collections.Concurrent;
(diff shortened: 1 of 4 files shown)


phroggster:
Trace data prior to `TraceLog.Init` is not captured. Will re-submit once solved.

PR #1334, opened 2017-10-02
Fill sys info from journal; add system cache

klightspeed:
* Move faction etc. info into the JournalLocOrJump base class
* Pull this info into the ISystem instance associated with the history entry
* Use this info if newer than systems DB


finwen:
will test and evaluate cache in master.

PR #1344, opened 2017-10-04
Avoid duplicate output in Star List Panel information summary.

andreaspada:
I added a check for water planet, so now the star list panel inform if the body is a "water planet" or a "terraformable water planet". Also, to avoid duplicate messages, the generic "terraformable" message apply only to all other bodies which are not water planet. It provide a cleaner display, especially for large star systems with lots of planets.


andreaspada:
Before:
starlist_before

After:
starlist_after

PR #1351, opened 2017-10-04
Fixed bugs reported in Discord

wokket:
- Reduced size of labels (apparently real-estate is a problem)
- Added transparency to images so toolbar works
- Tried not to make trnasparency too accurate so they work under any user-derived colour scheme (less successful)


EliteDangerous/JournalEvents/JournalScan.cs
```diff
@@ -487,13 +487,25 @@ public string DisplayString(int indent = 0, bool includefront = true)
             return scanText.ToNullSafeString().Replace("\n", "\n" + inds);
         }
 
+        public string GetHabZoneStringLs()
+        {
+            if (IsStar && HabitableZoneInner.HasValue && HabitableZoneOuter.HasValue)
+            {
+                return $"{HabitableZoneInner.Value.ToString("N0")}-{HabitableZoneOuter.Value.ToString("N0")}ls";
```
phroggster: The C# 6.0 interpolated string (which is typically handled by [`FormattableString`](https://msdn.microsoft.com/en-us/library/system.formattablestring(v=vs.110).aspx)) is a vast improvement over the olden days, but there's no need to use `{object.ToString("fmt")}` inside of one. Simply `{object:fmt}` is all it needs. In other words:
```CS
return $"{HabitableZoneInner.Value.ToString("N0")}-{HabitableZoneOuter.Value.ToString("N0")}ls";
```
is syntactically identical to
```CS
return $"{HabitableZoneInner.Value:N0}-{HabitableZoneOuter.Value:N0}ls";
```

EliteDangerous/JournalEvents/JournalScan.cs
```diff
@@ -487,13 +487,25 @@ public string DisplayString(int indent = 0, bool includefront = true)
...
+                    }
```
phroggster: I think the brace is getting attacked by space invaders! 👾  (trim the extra indentation, please)

EliteDangerous/JournalEvents/JournalScan.cs
```diff
@@ -487,13 +487,25 @@ public string DisplayString(int indent = 0, bool includefront = true)
...
+
         public string HabZoneString()
         {
             if (IsStar && HabitableZoneInner.HasValue && HabitableZoneOuter.HasValue)
             {
                 StringBuilder habZone = new StringBuilder();
-                habZone.AppendFormat("Habitable Zone Approx. {0}-{1}ls ({2}-{3} AU)\n", HabitableZoneInner.Value.ToString("N0"), HabitableZoneOuter.Value.ToString("N0"),
-                                                                                             (HabitableZoneInner.Value / 499).ToString("N2"), (HabitableZoneOuter.Value / 499).ToString("N2"));
+                habZone.AppendFormat("Habitable Zone Approx. {0} ({1}-{2} AU)\n", GetHabZoneStringLs(),
+                                                                                  (HabitableZoneInner.Value / 499).ToString("N2"), (HabitableZoneOuter.Value / 499).ToString("N2"));
```
phroggster: No changes needed, but to further show off the beauty of string interpolation:
```CS
habZone.AppendLine($"Habitable Zone Approx. {GetHabZoneStringLs()} ({HabitableZoneInner.Value / 499:N2}-{HabitableZoneOuter.Value / 499:N2} AU)");

EliteDangerous/JournalEvents/JournalScan.cs
phroggster: To further clarify, if either of the `double?` types are null, the two code blocks are not actually identical. `string s = $"{HabitableZoneInner.Value}";` will NRE if `HabitableZoneInner` is null, whereas `string s = $"{HabitableZoneInner}";` will yield an empty string.

wokket: Combination of trying to make the minimal change, and coding before 0600... good catch.

Both methods are protecting against null `double?`s already so I can't see the problem there.

phroggster: It's not a problem since the `null` checks are already there. I just wanted to correct my misstatement of "is syntactically identical to", to be "is *almost* syntactically identical to (unless `null` is involved)."

PR #1368, opened 2017-10-06
Remove unneeded `using EDDiscovery.DB` from tests

klightspeed:
`using EDDiscovery.DB` was causing compile failures on the EDDiscoveryTests project.


klightspeed:
Superseded by #1369

Issue #1317, opened 2017-09-29
Duplicate history entry


andreaspada:
Just noticed that my history is being duplicate.
Each and every system I enter is marked twice, and the visit counts is marked twice too...

> [2017-09-29 14:02:34Z] EDSM Fetch logs start with cmdr 3
> [2017-09-29 14:02:35Z] New entry 9/29/2017 5:02:00 PM Music
> [2017-09-29 14:02:35Z] New entry 9/29/2017 5:02:13 PM FuelScoop
> [2017-09-29 14:02:35Z] New entry 9/29/2017 5:02:31 PM StartJump
> [2017-09-29 14:02:37Z] New entry 9/29/2017 5:02:36 PM Music
> [2017-09-29 14:02:38Z] New entry 9/29/2017 5:02:00 PM Music
> [2017-09-29 14:02:38Z] New entry 9/29/2017 5:02:13 PM FuelScoop
> [2017-09-29 14:02:38Z] New entry 9/29/2017 5:02:31 PM StartJump
> [2017-09-29 14:02:38Z] New entry 9/29/2017 5:02:36 PM Music
> [2017-09-29 14:02:50Z] New entry 9/29/2017 5:02:50 PM FSDJump
> [2017-09-29 14:02:51Z] Arrived at system: Pyroifoea OJ-X d2-76 1:th visit.
> [2017-09-29 14:02:51Z] Arrived at system: Pyroifoea OJ-X d2-76 2:th visit.
> [2017-09-29 14:02:51Z] New entry 9/29/2017 5:02:50 PM Music
> [2017-09-29 14:02:52Z] New entry 9/29/2017 5:02:50 PM FSDJump
> [2017-09-29 14:02:52Z] New entry 9/29/2017 5:02:50 PM Music
> [2017-09-29 14:03:05Z] New entry 9/29/2017 5:03:05 PM Music
> [2017-09-29 14:03:13Z] New entry 9/29/2017 5:03:13 PM Scan

klightspeed:
That sounds like you are monitoring the journals using two separate directories (either differing only in case, or that are linked together).

phroggster:
Is it possible that the ED journal event was also duplicated somehow, or multiple instances of EDD were interacting, or the swamp gases reflected light from Venus? In other words, I don't think that we're doing anything wrong on our end, but do admit that it's possible even if it requires an obscure state to be hit.

[...]

PR #1382, opened 2017-10-10
Prevent runtime error when GetSystemsByName returns null

iainross:


iainross:
Don't merge this yet, it was fine in code but since I've built it I'm suddenly seeing https://github.com/EDDiscovery/EDDiscovery/issues/1317 happening...

iainross:
Deleted the last couple of journal files from my db and processed them again - it didn't recur.   Flew on for another bunch of systems against my local build and no recurrence.

I'm happy I've not introduced (or exposed) a systemic error but if anyone else wants to double check before this goes in that might not be a bad idea...

iainross:
Oh FFS.  It was just waiting for me to post that comment before it started happening again.  I need to track this down before this can go in really.

klightspeed:
Do the duplicate events have the same timestamp?  Need to check that it is properly matching the imported entries to the entries already in the history.

iainross:
From journalentries table for one offending system I've got a StartJump at 2017-10-11 00:09:58Z and an FSD jump at 2017-10-11 00:10:17Z which are in the file and all is good.  But a second FSD jump is being inserted with a TravelLogId pointing back at a dummy TravelLogUnit for jumps synced from EDSM.

Since it's reliable in build, patchy in debug and the behaviour suggests that the local and remote records are, at least briefly, not agreeing it looks like there's some kind of timing thing going on where we're pushing the jump to EDSM from one place and seeing it on EDSM from another before that other place has caught up with the fact that we did actually push it so it thinks it's syncing back from EDSM (why those wrongly synced from EDSM jumps have no coordinates even though EDSM knows them may be a separate issue or it may not)

Bed time here, I'll look at it more tomorrow unless someone else spots it in the interim.

klightspeed:
`GetSystemsByName` returning null means that the `api-v1/systems` API is returning an error.

<https://www.edsm.net/api-v1> does not mention anything about any rate limiting on the `api-v1/systems` API, though the X-Rate-Limit et al headers suggest it's 360 requests per hour.

PR #1384, opened 2017-10-11
Starlist tweaks

andreaspada:
I included a check in StarList Panel for moons, so now we now if a special body is a moon: terraformable moon, water moon, terraformable water moon, earth-like moon and ammonia moon; just in case...


andreaspada:
Closed, as need some refinements.

PR #1386, opened 2017-10-11
Enable code analysis, warnings, fix CS0675

phroggster:
* Enable code analysis for all builds. Now enabled for:
  * Debug|AnyCPU
  * Release|AnyCPU
* Enable warnings for all builds. Now enabled for:
  * Debug|AnyCPU
  * Debug|x64
  * Release|x64
  * Release|x86
* Disable visual studio hosting process for:
  * Debug|x64
  * Release|x64
  * Debug|x86
  * Release|x86
* x86 builds will prefer 32-bit code regardless of platform.
  * This will enable WOW64 if run on an amd64.
  * AnyCPU is still a better choice, but at least now the x86 build does what it says it's going to do on the tin.
* Fix a CS0675 warning in UserControls\UserControlSpanel.cs
  * `Bitwise-or operator used on a sign-extended operand; consider casting to a smaller unsigned type first`

### DIFF
--- a/EDDiscovery/EDDiscovery.csproj
+++ b/EDDiscovery/EDDiscovery.csproj
@@ -42,8 +42,8 @@
     <OutputPath>bin\Debug\</OutputPath>
     <DefineConstants>DEBUG;TRACE</DefineConstants>
     <ErrorReport>prompt</ErrorReport>
-    <WarningLevel>0</WarningLevel>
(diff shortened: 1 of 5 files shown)


phroggster:
I've split this up into bite-sized blocks to make it easier for auditing, and gone without several rearranges that were only distractions from the end goal. Future modifications will wish I had gone ahead with it, but oh well.

PR #1388, opened 2017-10-11
Avoid race resulting in duplicate entries

klightspeed:
Move the local history fetch to after retrieval of travel log from EDSM.  This should eliminate the race between the local history being fetched and the logs being fetched where new entries could be submitted and then retrieved.


iainross:
It stopped the duplication happening to me, looks good.

PR #1392, opened 2017-10-12
Added moon information to Star List Panel

andreaspada:
Now, the visited Star List panel tell us if a special body is a moon. 

A level is generated from ScanNode, which tell us if a body is a moon; than, both moon and non-moon bodies passes though the filter which tell us if they are worth mentioning: terraformable, water world, terraformable water world, earth-like or ammonia. 

It should work for procedurally-generated names as well for human-made names...


andreaspada:
Some examples:

starlist_moons
starlist_moons_b
starlist_moons_c

PR #1404, opened 2017-10-15
Small fixes in Habitable Zone information in SPanel

andreaspada:
I doubled the row height of the habitable zone information. It fixes this:
needfixforhabitablezone

Also, I changed the text in "Main Star Habitable Zone etc..." and "Others stars not considered"


iainross:
The same text is used in the body information of all the stars, so if you select a secondary star in the scan panel (or have it pop up when you just scanned it) then it will give you the hab zone of the secondary star but label it misleadingly as Main Star Habitable Zone.

You need to have a separate SystemHabitableZone string for SPanel purposes that doesn't make the per-body one confusing for stars that aren't the primary.

andreaspada:
You are right, I just noticed too. I just though in regard that... I'm closing for further refinements.

Issue #1406, opened 2017-10-16
Problem with keep-on-top and dialogue boxes


hansolovkarlsson:
If you turn on Keep on top under Settings, and then click Configure, the main app is on top, the dialogue is under, and you can't do anything. In Windows you can click and hold ctrl-tab to see the list of apps and windows, and then use the mouse to close the Configure dialogue to get out of it.

Not sure how this could be solved. Unless the dialogues somehow have an "on-the-most-top" automatically set. :)

phroggster:
There's a lot of places where this is an issue. Unfortunately, it's going to be hard to track down every instance of this, and it's even harder to properly notify every `Form` when it's `ParentForm.TopMost` property is being changed. The options are basically:
* Add an `OnTopMostChanging`/`TopMostChanging`/`OnTopMostChanged`/`TopMostChanged` event paradigm to `SmartSysMenuForm`, and ensure that everything that inherits from it adjusts any child `Form` instances that it's responsible for.
  * This is a lot of work to do correctly, but is the least obnoxious of the options.
* Add a similar event paradigm around `EDDConfig.Instance.KeepOnTop`, and have everything monitor that individually.
  * Would rather be able to have some windows topmost, and others not.
* Disable modifying `Form.TopMost` everywhere.
  * Easy to implement.
  * Not a fan.
* Deal with it as-is and the broken nature where a child `Form` instance could be hidden behind a top-most `ParentForm`.
  * ***Very*** easy to implement.
  * Not a fan.
* Modify every place that calls `Form.Show()` to set an initial `TopMost` value, but then never update it.
  * Easy to implement, but still a broken "*solution*".
  * Not a fan.

klightspeed:
Are the problem forms failing to set their owner?

phroggster:
You know... I was just testing that as well. Apparently we call `Form.Show()` 12 times without specifying a parent.

phroggster:
Wait a second... If we call `frm.Show(someOtherForm);`, then `someOtherForm` will never be allowed to obscure `frm`. I don't think parenting is the way to go in every instance.

phroggster:
After a pretty massive change touching quite a number of files and discovering a couple of unrelated bugs along the way, I'm proud to say that this has been merged in by way of 3e7535ac6b5138d2d88b07a22f7a6ab5c5aa4295. Expect this to be resolved in the next EDDiscovery release.

PR #1409, opened 2017-10-19
Fix Issue #1406

phroggster:
Essentially, replace every `Form.ShowDialog();` call with `Form.ShowDialog(otherForm);`, and correct the various places where we are using a `Control` instead of a `Form` (via `Control.FindForm()`).


phroggster:
An issue has been identified whereby `MessageBoxTheme` forms are displayed behind their `TopMost` parent Form. This is due to an out-of-place `CreateGraphics()` call in `MessageBoxTheme`, and is something that will need to be resolved before this can be merged. I'll work through this issue and reopen when it is ready for another go.

Issue #1441, opened 2017-11-05
There was an unhandled UI exception error on startup


SusannaSaunders:
No idea why it has started generating this error... Its worked fine up until now!
Here is the error log:

### Exception Details:
>There was an unhandled UI exception.
>```
>==== BEGIN ====
>System.ArgumentOutOfRangeException: Index was out of range. Must be non-negative and less than the size of the collection.
>Parameter name: index
>   at System.Collections.ArrayList.get_Item(Int32 index)
>   at System.Windows.Forms.DataGridViewRowCollection.SharedRow(Int32 rowIndex)
>   at System.Windows.Forms.DataGridViewRowCollection.get_Item(Int32 index)
>   at EDDiscovery.UserControls.UserControlEngineering.Display()
>   at EDDiscovery.UserControls.ChangedSelectionHE.Invoke(HistoryEntry he, HistoryList hl)
>   at EDDiscovery.UserControls.UserControlTravelGrid.FireChangeSelection()
>   at EDDiscovery.UserControls.UserControlTravelGrid.HistoryChanged(HistoryList hl)
>   at EDDiscovery.EDDiscoveryController.RefreshHistoryWorkerCompleted(HistoryList hist)
>===== END =====
>```

klightspeed:
Duplicate of #1435

PR #1442, opened 2017-11-05
Hook events after init is complete

### LINKED ISSUE #1435: Had a problem after EDDiscovery launched
### Description
<Wasn't really doing anything, just waiting for EDDiscovery to finish loading -->

klightspeed:
An async timer or callback could be called during any synchronous UI call.  Hooking events that can be called by async events can result in e.g. `Display` being called before init is complete.

Moving the hooks to the end of `Init` should fix #1435 and #1441.


robbyxp1:
Thanks you. PR is in

PR #1453, opened 2017-11-09
Give ButtonExt some attention

phroggster:
* `DefaultValue` attributes reduce *.designer.cs littering.
* Fully documented class and designer interface.
* Add support for the following properties in custom paint:
  * `AutoEllipses`
  * `RightToLeft`
  * `UseMnemonic`
* Mouse and keyboard interraction is identical to standard `Button`.
* Mimic the standard `Control.Paint` event even when custom painting.
  * Allows `Form`s, other `Control`s, etc, to customize further.
* Optimized requests for `Invalidate()`; only done if absolutely neccessary.
* More liberal disposals of `IDisposable`s.

Gratuitous side-by-side comparison:
buttonext1

### DIFF
--- a/ActionLanguage/ActionEditing/ActionPackEditorForm.Designer.cs
+++ b/ActionLanguage/ActionEditing/ActionPackEditorForm.Designer.cs
@@ -117,9 +117,6 @@
             // 
             // buttonInstallationVars
             // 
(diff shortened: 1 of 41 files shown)


robbyxp1:
Going to pull this in, buttonext, manually, so we don't get the 41 file diff stuff.

PR #1464, opened 2017-11-15
JournalScan conversions

dobb13:
JournalScan.cs: Put conversion factors in public consts. Update accuracy.
EDSMClass.cs: Use new public const in JournalScan.cs for Atmospheres -> Pascals


robbyxp1:
Ok, hold for a mo

PR #1506, opened 2017-11-27
Qohen Leth Icons + Icon Customization

klightspeed:
This makes almost every icon customizable


EDDIcons/IconSet.cs
```diff
@@ -0,0 +1,107 @@
...
+                            icons[name] = img;
+                        }
+                    }
+                }
+            }
+        }
+
+        public static Image GetIcon(string name)
+        {
+            return icons.ContainsKey(name) ? icons[name] : null;
```
phroggster: Probably fall back to `defaulticons` if `icons` doesn't contain it the key, and `null` only if both of those fail.

EDDIcons/IconSet.cs
```diff
@@ -0,0 +1,107 @@
...
+
+        public static void LoadIconsFromZipFile(string path)
+        {
+            if (File.Exists(path))
+            {
+                using (var zipfile = ZipFile.Open(path, ZipArchiveMode.Read))
+                {
+                    foreach (var entry in zipfile.Entries)
+                    {
+                        if (entry.FullName.EndsWith(".png"))
```
phroggster: I feel like a `ToLower` would be warranted here.

phroggster:
Also, and I'm talking from memory and not from thorough testing, but doesn't resgen strip file suffixes from resources names? It seems like loading from a zip file would include the suffix, yet the default icons wouldn't have it, leading to inconsistent naming schemes based on the icon load source.

klightspeed:
> Also, and I'm talking from memory and not from thorough testing, but doesn't resgen strip file suffixes from resources names? It seems like loading from a zip file would include the suffix, yet the default icons wouldn't have it, leading to inconsistent naming schemes based on the icon load source.

Embedded resources (accessed via `Assembly.GetManifestResource()`) do include the filename extension.

robbyxp1:
This is too big to check for 9.0.  The code is mostly stable as of now and I want to send it out on dec 16th.

Can we please pause until 9.0 is branched and out.

robbyxp1:
Note to me. Also, check out the interaction with the action packs - they use icons

andreaspada:
@klightspeed , I would like to test your commitment on my branch. IS it possible? Can I pull your commits to my local branch?

robbyxp1:
So, I've just tested it.

Correct me if I'm wrong, but:

The main problem is that it appears to replace all the current icons.  Which is not an acceptable solution.  We need to power up with the current set in place, and offer the option to go to another set.  I actually like the current icons, I'm sure other ones will.  They also have been tested in quite a lot of different themes.  If you can make it selectable in EDDIcons that would be good.

Its also lost the icon shading I did on some of them to adapt to different themes (I coloured the foreground)

I also have a big problem with using a 3rd party icons as the default set.  There will be more events coming into play in the next Q1 update. Someone has to draw them. We can't rely on him doing it in time.  

So, it needs to be reworked to 1) preserve the current set as the default 2) allow the other set to be swapped in 3) see if you can keep the icon shading for theme adaptation.

So i'm closing it for now.

klightspeed:
I will squash and use the original icon set by default.

robbyxp1:
Btw I think its nearly there, if you kept all the current icons in there, then had the new icons as alt_<journalname>, and selected between them on start up by a EDD config switch, it would be good.

PR #1511, opened 2017-12-02
Continue from the last system time when fetching systems

klightspeed:
EDSM Systems API will be limited to 1000 systems per request.  Take that limit into account, and continue fetching at the time of the most recently fetched system.

### DIFF
--- a/EliteDangerous/EDSM/EDSMClass.cs
+++ b/EliteDangerous/EDSM/EDSMClass.cs
@@ -270,8 +270,16 @@ namespace EliteDangerousCore.EDSM
                     break;
                 }
 
-                updates += SystemClassEDSM.ParseEDSMUpdateSystemsString(json, ref lstsyst, ref outoforder, false, cancelRequested, reportProgress, false);
-                lstsystdate += TimeSpan.FromHours(12);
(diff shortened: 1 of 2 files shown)


EliteDangerous/EDSM/EDSMClass.cs
```diff
@@ -270,8 +270,16 @@ internal long GetNewSystems(Func<bool> cancelRequested, Action<int, string> repo
                     break;
                 }
 
-                updates += SystemClassEDSM.ParseEDSMUpdateSystemsString(json, ref lstsyst, ref outoforder, false, cancelRequested, reportProgress, false);
-                lstsystdate += TimeSpan.FromHours(12);
+                long cnt = SystemClassEDSM.ParseEDSMUpdateSystemsString(json, ref lstsyst, ref outoforder, false, cancelRequested, reportProgress, false);
+                updates += cnt;
+                if (cnt < 100)
```
phroggster: Was that supposed to be 1000?

klightspeed: I originally had it as 1000, but I decided to reduce it in case the `LIMIT 1000` includes some rows that are filtered out.

PR #1519, opened 2017-12-08
Option to show current body mats in shopping list when landed

iainross:
Engineering/Synthesis/Shopping List panels are no longer historic (i.e. always use current mats whatever history entry is selected).

Shopping List now has context menu options to show all mats for current body or highlight available mats in the shopping list while landed.

Dismissing ship no longer sets IsLanded to false.

### DIFF
--- a/EDDiscovery/UserControls/UserControlEngineering.Designer.cs
+++ b/EDDiscovery/UserControls/UserControlEngineering.Designer.cs
@@ -46,15 +46,6 @@ namespace EDDiscovery.UserControls
             this.components = new System.ComponentModel.Container();
             this.dataViewScrollerPanel = new ExtendedControls.DataViewScrollerPanel();
             this.dataGridViewEngineering = new System.Windows.Forms.DataGridView();
-            this.vScrollBarCustomMC = new ExtendedControls.VScrollBarCustom();
-            this.panelButtons = new System.Windows.Forms.Panel();
-            this.buttonClear = new ExtendedControls.ButtonExt();
-            this.buttonFilterMaterial = new ExtendedControls.ButtonExt();
-            this.buttonFilterUpgrade = new ExtendedControls.ButtonExt();
-            this.buttonFilterLevel = new ExtendedControls.ButtonExt();
-            this.buttonFilterEngineer = new ExtendedControls.ButtonExt();
-            this.buttonFilterModule = new ExtendedControls.ButtonExt();
-            this.toolTip1 = new System.Windows.Forms.ToolTip(this.components);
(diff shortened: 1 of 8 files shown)


robbyxp1:
So, I don't like the removal of the historic nature, I like being able to see the situation at a point in the past.

Can we change this to an optional behaviour?  Should be easily done with a tick box..

ta.

phroggster:
I don't see any benefit for probably ~98% of users to be able to see historic material counts. It's both complicated and provides an unintuitive user experience to leave it as-is.

Maybe instead of a checkbox, a new UCCB, "Historic Material Information" (or something) should be added to provide something tailored for that, that can give one details about material counts, with charts and graphs and unicorns[1].

[1]: No actual unicorns, please.

robbyxp1:
I use it all the time to check back.  It also then not working like the other materials/commodities panels which obey the cursor.   We need to be consistent.  

And we should try not to remove functionality unless its obv. it needs removing for a good reason.

Please put the functionality back with a checkbox or something to say always current totals.

iainross:
There is already a Materials control that shows historic material counts, but if there's someone who actually does look at historic engineering/synthesis then I suppose I could make it optional.

Or I could make the stand alone panels travel history sensitive but not the shopping list?

iainross:
There was a reason for removing it - responding to the history cursor gives you as at the FSD jump into a system which means on re-loading the game on a planet the IsLanded didn't register until you took off and landed again or generated another journal by damaging the SRV or picking up mats.  I'm pretty sure that's also what was causing the values in the shopping list to sporadically revert to as-at-system-entry while prospecting.

robbyxp1:
Okay, thanks Iain.  Let me have a think.

robbyxp1:
The help pages still are saying that it responds to cursor pos.. thats one of the problems of just removing stuff or adding stuff, we have to remember to also update help ! ;-)

So I see you fixed the IsLanded.  if so, does not solve the issue above and therefore you could still leave it as tracking the cursor? With a tick to say ignore history position and just use last...

just asking.

iainross:
I don't want to update the help pages until its released :)

I did fix a couple of minor issues with IsLanded but if you've got the default filter of travel events on your History grid then you're correctly not landed at GetCurrentHistoryEntry on the cursor because that gives you the FSD arrival journal.

I've nearly got it configurable whether it tracks the cursor or not - just need to figure out why uctg.OnTravelSelectionChanged -= Display; isn't actually removing the event handler if I toggle it to off...

iainross:
Historicity is now optional...

robbyxp1:
Cheers, i'll merge in a mo..  real life interrupting.

PR #1532, opened 2017-12-12
New panel added, ScanGrid.

andreaspada:
Added a new panel, ScanGrid, which display the information of bodies scanned in a system in a grid, for an uncluttered and unobtrusive yet comprehensive view. Beside the body image, name and class, display some brief information, such as stellar mass and habitable zone, planets special state such as terraformable, ringed or volcanism and, for landable planets, it shows also the gravity.
Fields can be multiple selected, for easy copying and paste...

More improvement planned, such as row coloring (if possible), available materials list, rings composition and export, but the panel is already functional.

Modified the dataGridViewNearest name in the EstimatedValues panel to dataGridViewEstimatedValues, for better code understanding and naming convention; presumably, it was copied from the StarDistances panel.


andreaspada:
I added the Icon, which now successfully appear as intended, and manually fixed the Resources.resx path...
All tests passed.

klightspeed:
It seems VS2017 has a bad habit of lowercasing the name when inserting into the resx.  This breaks compilation when working on a case sensitive filesystem (as is the default in Linux and by extension Travis).

andreaspada:
@robbyxp1 , great addition to the panel with the bodies distances. However, the icon now do not show very well. I had choose a textured approach, so the icon were used at full size and squared, which gave good results. Now, they are circular, but with many glitches... it was something intentional?

andreaspada:
@robbyxp1 , please, dis-consider my last message. After a reboot, the icons shows very well and, frankly, they looks good rounded! ;)

PR #1533, opened 2017-12-13
Iconpack support

klightspeed:
* Crush all the PNGs to reduce their filesize - some small icons were hundreds of kilobytes in size
* Extract icons from the control local resx files
* Unify all of the window icons in EDDiscovery
* Organize the icons into area and function
* Enable loading icon packs from a zip file
* Use icons from the icon pack where possible


robbyxp1:
Hi,

Comments:

The EDDIcons and the IEliteIconSet in EliteDangerous looks good in themselves.

1. Why are all the original icons still in the EliteDangerous resource folder?  Why have they not been removed in favour of the ones in the icon pack?  Why are we now carrying two of the same icons.  This will make maintenance hard and confusing - are you editing the right icon etc?  Same goes for EDDiscovery resource icons.

2. The IconReplacer stuff in each class, and the need to call ReplaceIcons.  And the interaction with the designer and not being able to specify in it the icon pack directly.  All leaves me a little uneasy for new code and for new users trying to understand IIConPackControl,/EDDIconSet and IconSet.  Previously you just introduced the icon in the designer.  Is there a simpler way?

3. The Controls class in icon set - most of these members in the sub classes show No reference found.  Which is confusing.  Why have them if they are not referenced?  It seems to suggest that the icon is not needed. I see the swap functions using names "Toolbar.Open3DMap") instead of this controls reference.  One or the other method should be used, either remove the Controls bit where its not used or reference it in the swap..

robbyxp1:
Maybe we should just have empty icons in the designer, so you can remove all of the resource icons.

And maybe make it simpler to understand how the icons are to be replaced, the call back thru the iIconPackControl.. is that really needed? Why not just programatically pass the list of icons needed to be replaced.. (you may have done this due to control inheritance, but its still more confusing than a direct list)

robbyxp1:
More thoughts, if you had a placeholder icon in the resources (of each project), we would design with that, and then let the icon pack replace it programatically.

robbyxp1:
Thanks.. lets hold this until we have tidied up 9.0.0 and made a branch.  So say a week or so.

klightspeed:
Another merge fixup will be required after #1610 is merged.

PR #1536, opened 2017-12-14
Some improvements and additions.

andreaspada:
- Added atmosphere.
- Added rings counting, composition and inner and outer radius.
- Habitable Zone now shows in LS and AU.
- Distances now shows in LS and AU.
- Included EliteDangerousCore.JournalEvents for easier handling of constants and information math.
- The grid now Refresh on system changes and Update on new entries.


robbyxp1:
Thanks for the control.  I made even more improvements after this, a tooltip with the full info when hovered over the body name, and programatically drawing the image instead of using the image column (it does not allow fine control of positioning and size of an image). Also rows resize to show all content, which was a problem with the more data you added.

thanks

andreaspada:
Awesome, Rob. Many thanks! Those modifications you mentioned was perfect, I was thinking in the tooltip by myself, but it's still too difficult to me...

I'm very happy I can finally help and contribute with this project. Even if I'm quite limited by now, I hope I can improve my knowledge and understandings for more complex tasks, if needed.

All the best!

PR #1539, opened 2017-12-14
Added option for shopping list to indicate bodies in current system where mats are available

iainross:
Inclusion of EDSM scan data optional


robbyxp1:
Okay, it will be in 9.0.1

iainross:
Well darn, now I'll have to do a two step update to the documentation ;)

PR #1540, opened 2017-12-14
Minor tweaks and fixes to ScanGrid.

andreaspada:
@robbyxp1, thank you so much for your last improvements on ScanGrid. I'm very, very happy that finally I can contribute to this project.
I checked the last build, and noticed some fixes must be made, as well as some minor tweaks:

- Corrected the EstimatedValue routine call, to properly calculate post-2.2 values. If we call EstimatedValueED22 directly, as previously added, the bodies values are miscalculated.

- Tooltip with full information now shows also if hovering images. **Thanks you so much to add this!**

- Minor fixes in text spacing for stellar bodies and changed M: to Mass: (it can be confusing and, even if smaller, I found less easy on the eyes...).

- Transparent background under body images. (To prevent the first icon to always shows with the colored background...)

- Slight code commenting cleanup


robbyxp1:
Okay, its missed 9.0.0, but will be in the next one

robbyxp1:
Thanks.. 

Your not being defensive enough with nStellarMass.. it can be null so i've checked that. 

Presume all journal scan data can be not present.. Frontier are not consistent over time with what is filled.

PR #1544, opened 2017-12-16
Remove unnecessary CSharp 7 feature usage

mwerle:
Until it becomes necessary to use C# 7 features in the codebase, how about sticking to C# 6? (Apologies if this has already been discussed and decided upon)

This sort of syntactic sugar is not really necessary and stops people with older development environments from being able to compile the code and/or contribute.

### DIFF
--- a/EDDiscovery/UserControls/UserControlExpedition.cs
+++ b/EDDiscovery/UserControls/UserControlExpedition.cs
@@ -763,7 +763,8 @@ namespace EDDiscovery.UserControls
 
         private ISystem GetSystem(string sysname)
         {
-            SystemClassDB.TryGetSystem(sysname, out ISystem sys, true);
+            ISystem sys;
+            SystemClassDB.TryGetSystem(sysname, out sys, true);
             return sys;
         }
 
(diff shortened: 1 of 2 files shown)


mwerle:
These were the only changes which prevented it compiling on my computer (still on VS2015; will update to VS2017 when I get my new SSD)

mwerle:
Thank you :) And apologies if I came across short in the pull request - re-reading it it comes across rather entitled; that wasn't my intention.

phroggster:
No worries, and no apologies necessary.

We had a brief discussion about this maybe a month ago. The consensus was pretty much that devs were already moved over to VS2017, but that we'd try to avoid using the new lang spec features where possible until "*everyone*" had a chance to migrate over. Hence, you have my apologies for jumping the gun on it. Will try to keep my fingers away from the pattern matching (et. al.) for the foreseeable future.

mwerle:
Cheers mate.  
Well I'm not an active dev at the moment so if everyone is ready to move over don't hold back for me.  But some new language features don't add much - some even make readability worse imho!

Issue #1503, opened 2017-11-27
Searching in history .. slight histerisis before search to help with large sets


robbyxp1:
Basically, wait 1 etc second before filtering the travel history/journal.. so that it allows you to finish typing before applying the search

robbyxp1:
Done

PR #1555, opened 2017-12-19
Feat/search hysterisis

wokket:
Added slight delay after typing in search for #1503.


robbyxp1:
Holding due to wanting to have more time to test before i release it.

wokket:
I can understand wanting to test threading code out thoroughly!

I was trying to decide whether I needed a lock/double-check around the null checks/timer init code at https://github.com/EDDiscovery/EDDiscovery/pull/1555/files#diff-79af1cbabf12256f55df0afa0d690eceR255 and https://github.com/EDDiscovery/EDDiscovery/pull/1555/files#diff-79af1cbabf12256f55df0afa0d690eceR270.

And if so, whether moving the Dispose/null code onto the UI thread inside the Invoke removed that need.

I opted in the first instance to keep the code simpler and easier to reason about, and we can go from there.

robbyxp1:
Just trying to keep the master stable in the release cycle, its no comment on the code itself.  As soon as we have a stable 9.0.x i'll branch it and we can pull this in.

robbyxp1:
Replacing with PR1621, which uses a windows form timer instead - thats UI safe.

PR #1556, opened 2017-12-19
Some minor fixes to ScanGrid

andreaspada:
- Reordered some fields for better presentation of briefing information.
- Some minor punctuation and spacing improvement.
- Clear the Selection on draw, to avoid the first image to remain selected.


robbyxp1:
I'm not sure what you are editing the code in, but your stripping out excess whitespace at EOL which is causing extra diffs in the files.  Now I know they should not be there, but it does make comparison more difficult.

Thanks.

PR #1566, opened 2017-12-20
Correct gravity in UserControlScan

iainross:


andreaspada:
Yep, it was needed. I did it by myself, than throw out when saw that there is already a PR for that... ;)

PR #1567, opened 2017-12-20
Some addons and minor improvements to ScanGrid.

andreaspada:
Added a split panel to ScanGrid form for better layout management and total value summary.


robbyxp1:
Please rework this, like i did with the star distance grid, to use a panel at the top with the maxvalue, and the DGV below.  Splitter grids are only used when there is resizable content on both parts.

Use the Document Outline to arrange the panel/DGV in the order as shown in the Star Distance panel.

Set the panel to Dock:Top, and the DGV to Dock:Fill

put your controls in the panel.

Thanks

andreaspada:
Of course! It will be a pleasure!

PR #1569, opened 2017-12-21
Added StarsInRange panel and starsinrange panel icon

andreaspada:
A shameless copy of StarDistance panel, with the add of a minimum and maximum radius range. The systems are filtered and only those inside the defined constrains are display.

For my own use, but probably useful for someone else...


robbyxp1:
So, great idea.

Why don't you just add this functionality to the current nearest stars, with the default making it behave like the current one?

There are already 25+ panels.  To stop the number going up forever I'd rather you enhanced a current panel than create a new one..

Have a think.

andreaspada:
Of course, I can do that. I did not thought on that at first cause I feel that it could be useful to have BOTH panel, behaving each one its way. For example, the main use for the new StarsInRange panel, to me, is to rapidly see which stars already mapped are at jump range, check with my in-game route and avoid them if I want. 

But, obviously, i can do that with just a panel, opened twice and configured differently. 

I’ll give it a look.

andreaspada:
PR #1573 address this! Thanks

PR #1570, opened 2017-12-21
Added materials summary to ScanGrid and jumponium notification in StarList

andreaspada:
This add a materials list, formed by the element symbol and it's amount, to the ScanGrid panel.


andreaspada:
I modified the PR. I removed the list of materials, for better readability of the briefing, and added messages in case the body has materials for a level I, II or III FSD Boost, which is a by far a much more useful information to have at first sight.

andreaspada:
Removed the list of materials. Added a check in StarList panel which now notify bodies which has _jumponium_ materials. That bodies will display the contained material/s in the ScanGrid panel.

robbyxp1:
Suggest you merge back with the master, then resubmit.  Taking into account 1567

Thanks

PR #1571, opened 2017-12-21
Add a Compass panel to aid surface navigation

iainross:


andreaspada:
That sounds awesome!

EDDiscovery/PanelAndPopOuts.cs
```diff
@@ -60,6 +60,7 @@ public enum PanelIDs        // id's.. used in tab controls, and in button pop ou
             Trilateration,          // 28
             Settings,               // 29
             ScanGrid,               // 30
+            Compass,
```
andreaspada: Need the "// 31" for debug. ;)

EDDiscovery/UserControls/UserControlCompass.cs
```diff
@@ -0,0 +1,168 @@
...
+        private void Display()
+        {
+            if (last_he != null)
+            {
+                double? latitude = null;
+                double? longitude = null;
+                switch (last_he.journalEntry.EventTypeID)
+                {
+                    case JournalTypeEnum.Screenshot:
+                        JournalScreenshot js = last_he.journalEntry as JournalScreenshot;
```
andreaspada: What about adding a keys sequences to take "blind screenshots"? I mean, take a screenshot just for the purpose of get some coordinates, than purge it exactly after that... it could bypass the screenshot managing, to avoid mistakenly renaming, processing and movement of the same.

EDDiscovery/PanelAndPopOuts.cs
iainross: The whole point of an enum is that you don't need to know the numbers...

EDDiscovery/UserControls/UserControlCompass.cs
iainross: I did think about an 'auto-delete screenshots taken while I'm open' option and might add one later.

The coordinates come from the journal so it needs to be a screen shot taken in the game which we can't drive from EDD - we could have a control option to send the screenshot key press to the game and auto-delete the next screen shot but that feels like overkill to me.

EDDiscovery/PanelAndPopOuts.cs
andreaspada: Yes, I understand what yo mean, and I agree with you. However, is something was request by @robbyxp1 when I included my first panel, ScanGrid. So, I think is something's needed somewhat, or possibly only a requirement...

EDDiscovery/UserControls/UserControlCompass.cs
andreaspada: Well, indeed it's not a trivial task. Anyway, I'm very confident about that panel.

andreaspada: @iainross, maybe it could be awesome to have both proposed solution available to the user, which can select between on of them. The second one seems overkill but it could be the preferred approach for some user: if you are driving and find some interesting place to take a screenshot, and do not remember to close the compass, you'll find that you'll lose that magic moment... ,)

EDDiscovery/PanelAndPopOuts.cs
robbyxp1: Actually.. the reason i put the numbers in is because this is saved as an integer in the DB to mark the panel, hence i needed to know the number of the enums ;-)

PR #1573, opened 2017-12-22
Added radius to Nearest Stars panel and duplicates removal

andreaspada:
Modification of StarDistance panel with a an inner and outer radius definition, and a checkbox to switch to the former or new behavior. Whenever the checkbox or the textboxes are modified, the panel refresh itself.
The controls reside in a split panel, and they can be completely hidden as well as re-position itself for narrower sizes.

It replace the StarInRange panel attempt, to avoid one more panel with similar functionalities. Thanks @robbyxp1 for your feedback!

stardistance-mod

### DIFF
--- a/EDDiscovery/UserControls/UserControlStarDistance.cs
+++ b/EDDiscovery/UserControls/UserControlStarDistance.cs
@@ -95,15 +95,52 @@ namespace EDDiscovery.UserControls
             SetControlText("");
             dataGridViewNearest.Rows.Clear();
 
+            double minRadius;
+            double maxRadius;
+
+            // handle empty textboxes, assign default values or user defined ranges
+            if (textMinRadius.Text != "")
+            {
+                minRadius = double.Parse(textMinRadius.Text);
+            }
+            else
+            {
+                minRadius = 0;
+                textMinRadius.Text = minRadius.ToString();
+            }
+
+            if (textMaxRadius.Text != "")
+            {
+                maxRadius = double.Parse(textMaxRadius.Text);
+            }
+            else
+            {
+                maxRadius = 20;
+                textMaxRadius.Text = maxRadius.ToString();
+            }
+
             if (csl.Count() > 0)
             {
                 SetControlText("Closest systems from " + name);
                 foreach (KeyValuePair<double, ISystem> tvp in csl)
                 {
                     int visits = discoveryform.history.GetVisitsCount(tvp.Value.name, tvp.Value.id_edsm);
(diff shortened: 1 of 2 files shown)


robbyxp1:
Perfect..

andreaspada:
Thanks you so much, @robbyxp1!

During I was working on the mod, I thought that possibly a third textbox to define a broader or narrower maxitems could be interesting. Do you think it's something worth or useful enough? Currently, it's hardcoded to 50... near the core, there are hundreds of systems in a very small radius, so for special needs could be useful have some more "radar" sensitivity...

andreaspada:
I added a code which hopefully solve the duplicated entries issue: #1579

EDDiscovery/UserControls/UserControlStarDistance.cs
```diff
@@ -137,10 +138,44 @@ private void FillGrid(string name, SortedList<double, ISystem> csl)
...
+                    // Check for duplicated entries and remove them
+                    if (dataGridViewNearest.Rows.Count > 2)
+                    {
+                        for (int currentRow = 0; currentRow < dataGridViewNearest.Rows.Count; currentRow++)
+                        {
+                            var rowToCompare = dataGridViewNearest.Rows[currentRow]; // Get row to compare against other rows
+
+                            // Iterate through all rows 
+                            //
+                            foreach (DataGridViewRow row in dataGridViewNearest.Rows)
```
wokket: I'm not sure how many rows we're expecting in this grid, but this inner loop only needs to look _forward_ from `currentRow + 1` inclusive.  Any rows between 0 and `currentRow` have already been checked against this row, and can't be a dupe.

This also simplifies some of the logic slightly by removing the need for the equality check.

andreaspada: Thanks, @wokket . I'm currently traveling, so unable to work on that. If you have a working alternative, and if you are able to include it, it will be great!

wokket: I haven't tried to update someone else's PR before, but I'll have a look today.... worst case I'll submit a change once this approved if you like.

robbyxp1:
I'm re-engineering it in case you see it.. you should use a panel at the top, set to dock:top, and the DGV below, set to fill, and it needs to be in the right order in the document outline view to work.

This is a exception waiting to happen :             minRadius = double.Parse(textMinRadius.Text);

It will except if text is not valid..
use textMinRadius.Text.InvariantParse

One of our extension methods (see BaseUtils)
which then reads:

          double minRadius = textMinRadius.Text.InvariantParseDouble(0);
            double maxRadius = textMinRadius.Text.InvariantParseDouble(20);

Calling Init() again is not allowed.

I've reworked it in another PR.

PR #1575, opened 2017-12-26
Improved performance opening 3D map

iainross:
Remove checking of all travel history with no coordinates against EDSM before opening 3D map
Ensure that a history refresh triggered by updates to EDSM systems updates travel history that now has coordinates


andreaspada:
Great!

PR #1592, opened 2018-01-05
Correct PR1575 and enhanced

robbyxp1:
Now saves min/max to DB
SystemClassDB supports min/max distances for searching
HistoryList supports min/max for searching
UserControlStarDistance - removed duplicate removal. Uses new lookup
interfaces.  Prevent assert


andreaspada:
Great! Many thanks!

PR #1595, opened 2018-01-05
Total value and materials in ScanGrid. Jumponium notication in StarList.

andreaspada:
Rewrited following the directives of @robbyxp1.

Added jumponium notification to StarList: it shows the body names, and how much of each FSD boot level it has.

Added materials for jumponium in ScanGrid: for each body which has jumponium material, display which material is available.

Added total scans value to ScanGrid: it calculate the approximate total value of the scanned bodies in the system.

### DIFF
--- a/EDDiscovery/UserControls/UserControlScanGrid.cs
+++ b/EDDiscovery/UserControls/UserControlScanGrid.cs
@@ -47,13 +47,15 @@ namespace EDDiscovery.UserControls
             // this allows the row to grow to accomodate the text.. with a min height of 32.
             dataGridViewScangrid.DefaultCellStyle.WrapMode = DataGridViewTriState.True;
             dataGridViewScangrid.AutoSizeRowsMode = DataGridViewAutoSizeRowsMode.DisplayedCells;     // NEW! appears to work https://msdn.microsoft.com/en-us/library/74b2wakt(v=vs.110).aspx
-            dataGridViewScangrid.RowTemplate.MinimumHeight = 32;            
+            dataGridViewScangrid.RowTemplate.MinimumHeight = 32;
+            this.dataGridViewScangrid.Columns["ImageColumn"].DefaultCellStyle.SelectionBackColor = System.Drawing.Color.Transparent;
         }
 
         public override void Init()
         {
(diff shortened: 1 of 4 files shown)


robbyxp1:
Cheers, will merge today or tomorrow, dep on my availability

andreaspada:
@robbyxp1, I saw that you already created the page in the wiki for the ScanGrid panel, which is great. I would like to contribute a more detailed explanation, is it possible to have edit permissions for that page?
This can be important later, when I'll be able to implement more functionalities - such as a couple of button toggles - for better management of information, so I could include updated information on the wiki...

EDDiscovery/UserControls/UserControlScanGrid.cs
```diff
@@ -226,10 +272,12 @@ public override void InitialDisplay()
...
                         DataGridViewRow cur = dataGridViewScangrid.Rows[dataGridViewScangrid.Rows.Count - 1];
 
-                        string scan = sn.ScanData.DisplayString(); // display tooltip with full information when hower bodies image and name
+                        string scan = sn.ScanData.DisplayString(); // display tooltip with full information when hower bodies image
                         cur.Cells[0].ToolTipText = scan;
-                        cur.Cells[1].ToolTipText = scan;
                         cur.Tag = img;
+
+                        // display total scan values
+                        BuildSystemInfo(last_sn);
```
robbyxp1: This build is in the wrong place, you don't do it ever node iteration. fixed on master

robbyxp1:
Comments..  Don't use LabelExt, its for special purposes only.  Just use Label.

You can put the total in the control text instead of using valuable space on the screen, i've done that on the tip.

Thanks for the push.

robbyxp1:
Also, a lot of your pushes are showing up line deltas where only the space is removed.  Now I know its better, but it generates more stuff to check.  I'm not sure what editor your using but please disable the space removal at end of line option.

EDDiscovery/UserControls/UserControlScanGrid.cs
andreaspada: Great, thanks to let me know and help in understand. ;)

PR #1598, opened 2018-01-07
StatsPanel mod suggestion.

andreaspada:
Some modifications on the StatsPanel: moved graphs in a separate tab (I'm planning to add more graphs).
Separated current system from global statistics, and added some information and useful distances.

@robbyxp1, all the work is mainly a preparatory step for the addition of a search tab, to let the user query the db for bodies name, classes and properties.


andreaspada:
@robbyxp1, I'm also working on a RadarPanel, which should provide a neighborhood scatter 3d plot of the nearest stars. I don't know OpenTK at all, so I was thinking in use a 3rd party, free, cross-patform scatter chart library, but I'm pretty sure as we already have OpenTK it should be better to use it. Could you kindly help me in understand how to initialize an orthogonal 50Ly cube, so I can proceed with the remaining code?

robbyxp1:
The best I can suggest is to look at the 3dmap code.. that's how i learned opentk/opengl.  its not a trivial task!

robbyxp1:
Need to fix up the graph - its not refreshing when first opened (appears white) only when you click on a travel entry does it update..

otherwise, okay.

andreaspada:
That's ok! Thanks for check that. I'll work on that later at night... it's not something urgent, as I said, is mainly a preparatory task.

PR #1600, opened 2018-01-07
Radar panel - NOT FOR MERGE - PR ONLY FOR SHARE INSIGHTS

andreaspada:
@robbyxp1, this is what I made so far. I just created a "pseudo-wroking" mockup, using random generated coordinates to simulate the coordinates which I'm still unable to retrieve. I have not made any real coords calculation, too; I just fiddled around a played with some ideas...

The checkbox of course is temporary. Having some different angles to view from can be interesting, especially for multiple opened windows...

I personally think that a 3D rotatably view should be the really best, but a 2d plot could have its strengths too (not too weight on the GPU), so I prefer to implement both solutions, if you main developers do not mind.

I also noticed, during use, that the last modification you made to the StarDistance panel suffer from a problem I noticed, and that why I opted for the checkbox to enable/disable the filter. I pretend to have two simultaneous windows opened, one filtered, small sized, the other with the full list. The way the share the same ranges, now, prevent that. Do you mind to re-enable a pass-thrue switch? Maybe, when the checkbox is unchecked, the min and max radius could hide from view, to save some screen space...


andreaspada:
@robbyxp1, I added a pseudo3d chart, for test and fun. As multiple series, in 3D points charts, are rendered as layers in perspective, the idea is to have x and z coords plotted as points, as usual, while the y coord define a serie. So, if we have a maximum range of 50ly, we have a span of -50 to 50 (100) ly for the y coord, so each ly represent 1%. If we have 25 series, they represent 4% each, so every 4ly we have a step through the series sequences.

robbyxp1:
They should not share the same ranges.. i'll check

robbyxp1:
just checked, the star distance panels have their own saved values, i had thre open all with different values.  So I don't want that check box back.

andreaspada:
> just checked, the star distance panels have their own saved values, i had thre open all with different values. So I don't want that check box back.

@robbyxp1, that's confusing... I have two panel opened; if I set one with a min and max radius as, said, 30 and 70, and the other as 0 and 1000, if I close the app and open it they return both with 0 and 1000 as range. There's a way to have them store the values as by panel instance? Is it something possible to achieve?

phroggster:
`DbSave` (from `base.displaynumber`) is your per-instance save/load identifier. Each instance will (or at least, is supposed to) have a different `displaynumber`, hence a different place to read/write configuration.

Probably the easiest thing to do is look at all of the `displaynumber` references throughout the codebase to see how to use it in practice.

phroggster:
Also, suggest rebasing onto master to fix the PR conflict. I know this isn't intended for merging, but it makes things easier to keep it up to date. 0c10fa7be384dd399fd4e409d61b92f98b093d4f really did a number on that file.

I haven't done anything with it in a while, but at one point I created a [modern-day OpenGL galaxy browser](https://github.com/phroggster/phrogGLControl) using the EDDiscovery systems database. If you really want to migrate towards actual 3D drawing routines, it may be worth a peek (if it even compiles any more).

andreaspada:
Thanks, I'll do a rebase ASAP. Can you help me understand how to pars the coordinates from the nearest list  to the FillRadar method? I still didn't understand how, and I'm quite advancing with the other things...
By the way, I already have some working rotation with mouse in the 3d plot, yet needs more work.

phroggster:
Something like:
```CS
double MaxRadius = 75;
double MinRadius = 0.0;
private void KickComputation(HistoryEntry he)
{
	if (he != null)
	{
		double curX = he.System.x;
		double curY = he.System.y;
		double curZ = he.System.z;

		computer.CalculateClosestSystems(he.System,
			(s, d) => BeginInvoke((MethodInvoker)delegate { NewStarListComputed(s, d, curX, curY, curZ); }),
			maxitems, MinRadius, MaxRadius);
	}
}
private void NewStarListComputed(ISystem sys, SortedList<double, ISystem> list, double curX, double curY, double curZ)      // In UI
{            
	System.Diagnostics.Debug.Assert(Application.MessageLoop);       // check!

	discoveryform.history.CalculateSqDistances(list, sys.x, sys.y, sys.z, maxitems, MinRadius, MaxRadius);

	FillRadar(sys.name, list, sys, curX, curY, curZ);
}
private void FillRadar(string name, SortedList<double, ISystem> csl, ISystem sys, double curX, double curY, double curZ)
{
	SetControlText("");

	if (csl.Count() > 0)
	{
		SetControlText("Plot of closest systems from " + name);
		foreach (KeyValuePair<double, ISystem> tvp in csl)
		{
			if (sys == tvp.Value || tvp.Key > MaxRadius || tvp.Key < MinRadius)
				continue;
			// etc.
		}
	}
}
```
Plus updating `MaxRadius` and `MinRadius` whenever the appropriate `TextBox.Text` values change (and saving/restoring these `double` values directly instead of the string interpretations).

PR #1601, opened 2018-01-08
Apply themeing to TabStrip sub-controls

phroggster:
[113a8314](https://github.com/EDDiscovery/EDDiscovery/commit/113a83144532657195e88c6c26a43c8535559d64#diff-a656338ecbcf571e60eeffc8f54125cdR814) seems to have added a return that wasn't needed?

Previously, changing theme during runtime would yield mixed colouring for the UserControlHistory tab, as it makes extensive use of TabStrips inside the SplitContainers. Now, all of the splits get updated appropriately.

However, the UserControlScan control may still have odd fonts and scroll bar, depending on the image size, scroll position, and theme choices. Better than it was, though! (Would have to migrate `ThemeStandard` back to `EDDiscovery` namespace to be able to correct this fully. Too much work...)

Here's before and after shots. Both images were taken starting up in the Windows Default theme, then switching over to the Elite Verdana theme.
Before: 
01-before

After:
02-after

### DIFF
--- a/ExtendedControls/Themes/ThemeStandard.cs
+++ b/ExtendedControls/Themes/ThemeStandard.cs
@@ -818,7 +818,7 @@ namespace ExtendedControls
                     }
                 }
             }
-            else if (myControl is TabStrip )
+            else if (myControl is TabStrip)
             {
                 TabStrip ts = myControl as TabStrip;
                 ts.DropDownBackgroundColor = currentsettings.colors[Settings.CI.button_back];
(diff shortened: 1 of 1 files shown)


robbyxp1:
Return is put in normally when we don't want the theme its sub components.. So that is why it would be there.. Did you check all the derived classes of TabStrip?

robbyxp1:
Otherwise it looks okay.. probably bring in anyway, and hope for the best.

phroggster:
This has been bugging me for some time. Can't believe that it took me this long to figure out what was going on.

I see 32 `TabStrip` references, with nothing inheriting from it, and only `UserControlHistory.designer.cs` owning any of them.

PR #1604, opened 2018-01-08
Add app version info to the ExceptionForm

phroggster:
* Both in the dialog Text and the Markdown-formatted clipboard details.


robbyxp1:
Great idea.. funny enough i was wondering about that too.

phroggster:
I had meant to include this initially, as it's very helpful information, it just somehow never came to be. It's amazing how the significant details can get overlooked.

PR #1606, opened 2018-01-09
Writing to EDSMID/JSON pos now in a big transaction for speed.

robbyxp1:
Improves performance of scrolling thru old stuff

### DIFF
--- a/Audio/AudioDriverCSCore.cs
+++ b/Audio/AudioDriverCSCore.cs
@@ -122,8 +122,11 @@ namespace AudioExtensions
 
         public void Dispose()
         {
-            if ( aout != null )
+            if (aout != null)
+            {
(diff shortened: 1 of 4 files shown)


Audio/AudioDriverCSCore.cs
```diff
@@ -122,8 +122,11 @@ private void Output_Stopped(object sender, PlaybackStoppedEventArgs e)
 
         public void Dispose()
         {
-            if ( aout != null )
+            if (aout != null)
+            {
+                aout.Stop();
                 aout.Dispose();
+            }
```
phroggster: Should probably nullify `aout` after disposal to prevent `aout.Stop()` from throwing if this were to get called multiple times.

PR #1610, opened 2018-01-10
Added two panels: Plot and Radar.

andreaspada:
This PR adds two more panels: Plot, which plot in 2D the nearest systems from three point of view (front, top, side); and Radar, which shows a 3D map, also of the nearest systems. Both display systems inside a radius/shell, user defined and which is saved in the db in two distinct variables, so they can have different default values.

**How Does it Works**

**Plot**
The Plot are Bubble charts, which maps two coordinates as X and Y axis, and the third coordinate as bubbles sizes, so to create a fake perspective: the nearest systems are larger: the farther, smaller.
Visited systems are shows in blue, non visited in yellow. Current system is red.
Axis are resized to the maximum radius, whenever it changes.

**Radar**
The Radar use a 3D Point Chart, with X and Z coordinate mapped as X and Y axis, and the Y coordinate is mapped through 101 layers of series. The current system is mapped exactly in the middle of the series, so there are 50 series before and after the current system, available to be mapped. As soon as the coordinates and axis distances are calculated, the Y distance is normalized to fit inside the span of the layers of series, and mapped to the series which fits best.
The series depth is adjusted to the maximum radius span, so to have a more true orthogonal view. A slight perspective is applied.
Axis are resized to the maximum radius, whenever it changes.
The Radar can be freely rotated with the mouse.

The Radar has working tooltips which shows the name of the system and its distance from the current system.

---

Both needs more work, but I would like to have some developer and power user to test it during live use, to evaluate the usability and possible tweaks and improvements. They seems perfectly usable, but needs more refinement.

**Known Bugs**
- The Plot also has tooltips coded, like the Radar, but for some reasons they do not work anymore after I added two additional views. They works if only one charts is loaded, though... any clue?
- Radar view sometime insists in put more systems in the "current system" series. I didn't understand if they are "ghosts" current systems, not cleared, or systems which are wrongly mapped as current.
- Plot panel crash when opened as popup. Works perfectly if opened in the grid...


robbyxp1:
Ok, its in the pipeline for adding.. may be a few days, may want to do a 9.0.3 first before we add anything else

andreaspada:
@robbyxp1, if you have better names for the panels, do not be shy! ;)

andreaspada:
@robbyxp1: **Planned improvements:**

- Add a mouse menu to the systems - both in Plot as well in Radar - to: open them in EDSM, set as target
- User definable colors

robbyxp1:
Is this ready to be incorporated? I'm not sure on your position.  You said in chat you wanted to improve it more.  Rob

andreaspada:
Well, I will not have many spare time to work on that in the next days. I think it’s better to incorporate to let anyone experiment with that, so may appear ideas and improvements from more experienced coders. What do you think?

robbyxp1:
I think we will wait.. we need things to be reasonably mature before we expose it to our audience, otherwise we get a multitude of Problem reports!

I'll kill the PR for now, just reopen it when you think its more mature.

andreaspada:
That's ok. Indeed, I still need to solve the known bugs. Btw, have you some clue why the Plot crash when opened as popup while the Map opens without problems? They behave quite similar, I haven't figure out...

PR #1614, opened 2018-01-11
Add portable zip file to app update dialog

### LINKED ISSUE #1611: Regression: Update dialog no longer shows button to download portable version
The update dialog used to have a button to directly download the portable version, in the same way it now has a button for the installed version.

phroggster:
Fixes #1611

This went overlooked when the portable zip was renamed for 8.3.3.

Squashed a number of other bugs discovered along the way:
* Match GitHub file names case-insensitively.
* `NewReleaseForm` dialog can now be resized.
* Fix `Form` min/max sizes being ignored by `DraggableForm`.
  * Was previously allowed to get down to `2x2`...
  * `Form.MaximumSize` and `Form.MinimumSize` values are now respected.
* Stop <kbd>Alt</kbd> keypresses (with no secondary key) from opening the system menu.
  * Only show the system menu with <kbd>Alt</kbd>+<kbd>Space</kbd>.

### DIFF
--- a/BaseUtils/HTTP/GitHubRelease.cs
+++ b/BaseUtils/HTTP/GitHubRelease.cs
@@ -52,7 +52,7 @@ namespace BaseUtils
         {
             get
             {
-                var asset = jo["assets"].FirstOrDefault(j => j["name"].Str().EndsWith(".exe"));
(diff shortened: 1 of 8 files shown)


phroggster:
Some screenshots of this in action, since no one will see it for another two releases (one to get the fix, and another to see the update notification):
windefault
orange_delight

PR #1616, opened 2018-01-11
RollUp for some Panels toolbar.

andreaspada:
Scan, StarList and StarDistance now has a RollUp panel, which can pinned or unpinned.
Added a RollUp panel to ScanGrid, which display the approx total scan value.


robbyxp1:
I'm probably going to fix it another way, so this will probably not be needed... thanks

robbyxp1:
I've added control text/title to UCCR, to make it orthogonal with the rest of the system, so this is not needed.

May be better next time to flag it up rather than code around!  Saves you work. Thanks for doing it though.

Rob

andreaspada:
No problem, for me any work is good. I’m not a professional coder, so any chances to understand how to achieve things are welcome.

Anyway, I will stay with some roll up here and there (i.e. StarDistance) to avoid unnecessary screen space use. Or, maybe worth a think, migrate the most possible controls to title bar.

PR #1618, opened 2018-01-12
Limit DraggableForm Caption Mouse Button Events

phroggster:
As discussed on Discord:

* Add a `uint CaptionHeight` property to `DraggableForm`.
* Limit caption mouse down/up events to within this value.

### DIFF
--- a/ExtendedControls/Forms/DraggableForm.cs
+++ b/ExtendedControls/Forms/DraggableForm.cs
@@ -36,6 +36,16 @@ namespace ExtendedControls
         }
 
 
+        /// <summary>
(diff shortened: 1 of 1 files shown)


phroggster:
As noted on Discord, I'm not a fan of this approach, but it is better than the way things were. The *right* solution is to only flag controls as being caption controls if they are actually caption controls. The `NewReleaseForm` shows a good way of handling this, with a dedicated `Panel pnlCaption` control that contains various caption sub-controls. However, the cost of implementing this throughout would be quite a bit.

robbyxp1:
Yep, its a good bodge for now

PR #1622, opened 2018-01-13
Redo DrawnPanel

phroggster:
Lots of changes going on here, but it behaves better in all scenarios.
* Inherits from `Control` instead of `Panel`. No more controls disappearing into it.
* Supports `IButtonControl` to behave just like a button.
  * `Click` is appropriately handled to catch keyboard and/or **left** mouse.
  * `DoubleClick` isn't supported. Sorry, but no need. Could add if/when desired using a `Timer`.
  * `MouseClick` is no longer mandatory, but is still supported.
* Mnemonic and right-to-left language support for text.
* `DefaultValue` attribs and full commentary specified throughout.
* Split decendant types off into individual files.
* Invalidates appropriately where it should.
* Add `ImageSelectedChanged` event.
* `ImageSelected` painting routines tested over large range of sizes. Only the `TransparentClickThru` and `FullyTransparent` are weak at large sizes.
* Property changes:
  * `ImageText` renamed to `Text`. base provided.
  * `BackgroundImageLayout` prop can control `Image`. base provided.
  * `DrawnImage` prop renamed to `Image`.
    * Drawn in background, meaning that `Text` or `ImageSelected` *can* be drawn over it.
  * `MarginSize` prop moved to `Padding`. base provided.
  * New `Selectable` prop allows to not grab focus from clicks.
  * New `DialogResult` prop. Yes, we really can be a button!


robbyxp1:
26 files.. too big for tuesday release.. hold for next one

robbyxp1:
I may change my mind of course ;-)

phroggster:
When I get back to my desk in a little bit, I'll split off a PR for the unrelated stuff noted at the bottom.

Most of to "26" are designer files, and/or switching from `MouseClick` to `Click` (`Click` didn't previously function *correctly*, getting called for right-clicks, etc.). Up to you if you want to sort through it all before 9.0.3 or not, but the errata fixes should definitely get added in sooner.

phroggster:
I've split the less-related stuff off into #1624, which should be merged in prior to this to minimize conflict.

PR #1627, opened 2018-01-14
Fill name and name in EDSM log fetch

klightspeed:
This should fix the rare case where the system name is null when retrieving logs from EDSM.


klightspeed:
Superseded by #1632

PR #1637, opened 2018-01-17
Fixup 1622

phroggster:
Some changes in #1533 weren't compatible with #1622. Fix that up.


robbyxp1:
Been there, done it ;-)

PR #1638, opened 2018-01-18
Added Pan and Zoom to MapPanel - Preliminary Work!

andreaspada:
- Added pan movement when middle button is pressed.
- Added zoom using mouse wheel. 

The Zoom is a dirty hack; MSChart do not have zoom neither pan when display charts in 3D, so I had to rely on another method. I created an invisible pictureBox, which give references sizes without overlap with the panel (may change). Than, I undocked the chart and let it position itself relative to the picturebox, to stay reasonable in the center (need more refinements). When the mouse whell is scrolled, the chart changes in size and recalculate its location to remain in center. 

If the control is resized, the chart is redefined with default values...

---

Known Needed Refinements:

- mouse rotation still need to follow better the mouse position. I tried, but no success yet...
- pan should be smoother
- zoom maybe should center under the mouse position?


phroggster:
* The chart doesn't appear to be centered in the control without panning.
* Panning induces epileptic seizures, as the grid lines flicker extensively.
  * It doesn't start as soon as middle mouse button is depressed, but as soon as the cursor is moved one pixel in any direction, then goes nuts.
* Zoom doesn't appear to follow the cursor. It seems to pick a random spot?
* All sorts of triggers cause the orientation to get reset.
* Check out #1640/#1641 for a way to **not** continuously parse the two `TextBox.Text` values.
* Also noted in #1640/#1641, `computer.CalculateClosestSystems` gets all known systems from the DB. In theory, this should include all known systems from the history, rendering the call to `discoveryform.history.CalculateSqDistances` (and the entirety of `NewStarListComputed` after the `TextBox` continuous parsing is removed) moot.

Aside from those issues, it looks like it could be pretty nice. I've yet to play around in earnest with `System.Windows.Forms.DataVisualization.Charting`, so I really have no idea what it's capable of, but it seems like you might be on to something that many people would prefer over the current 3D map interface.

andreaspada:
Thanks for the comments, @phroggster. I'm aware of almost of those issues. It's why I asked for help, btw. I'm not an experienced coder, I'm still learning while getting my hand dirty, so I follow many direction until I found something that work, which is not so bad, because at least it give me enough chances to experiment with various features of c# and visual studio. However, it means that I do not follow a premeditated, well based thought, which means sometime I have no clues why something work and why something do not work. Sorry! ;)

But, I'm doing my best. 

I'll definitely have a look at #1640 and #1641. They are very welcome!

By now, I mainly coded a kind of a working mockup for a "can be good if did the right way" idea. I'll try to address all the issue one at a time. If you have some code snippet which can help, let me know.

And, really, thank you so much for your time. I'm really happy to be finally able to contribute, even if for such minor things, to this awesome project. All the best!

andreaspada:
Just noticed that using the Control.PointToClient method() I can solve at least the rotation bug.

PR #1640, opened 2018-01-18
Fix UserControlStarDistance

phroggster:
* Squash a bug that replaced all stars with only visited stars.
* Don't re-parse the text values for every change.
* Set the DataGridView columns read-only. No sense allowing people to rename systems...
* Fix a few potential NRE encounters.
* Update the TextBoxes if the parse result isn't valid.
* Filter out the current system instead of overriding the minimum value.


robbyxp1:
Are you saying the 9.0.x version crashes..  If so it should also be patched.

robbyxp1:
See my discord comments

PR #1651, opened 2018-01-20
Some UserControlStarDistance Enhancements

phroggster:
* Don't re-parse the text values for every change.
* Set the `DataGridView` columns read-only. No sense allowing people to rename systems...
* Fix a few *potential* NRE encounters.
* Update the `TextBox`es even if the parse result isn't valid.
* Filter out the current system instead of overriding the minimum value.
* Allow numeric sorting on number of visits column.

From #1640, with updates for the comments, and a rebase onto the 7befdcf8d5ff3ebc1728f2c2d3353cadc2d4b407 fix for a clean merge.


robbyxp1:
My next mega push is going to clash with this.. so i'll have to merge.. note to Rob.

PR #1655, opened 2018-01-26
More DrawnPanel Enhancements

phroggster:
* Draw focus rectangles when appropriate.
* Pass the `MouseEventArgs` from `OnMouseClick` to `OnClick`.
* AntiAlias almost all of the painting.
* Inset the Ross logo so it's more visible.
* Fix inverted background colours to match non-inverted foreground.
* More commentary throughout.
* Throw an exception if `ImageSelected` wasn't case'd during paint.
  * Will catch problems if the `enum` ever grows without painting support.
* Set `UserControlSysInfo` `DrawnPanel` control `Padding` values appropriately.
  * Mimics previous appearance, prior to the last `DrawnPanel` update.


robbyxp1:
;-)

PR #1657, opened 2018-01-26
EDSM Data control, 2d/3dmap

robbyxp1:
EDSM Data can be controlled/turned off with a selectable galaxy map.
2dmap themed and improved functionality
3dmap themed
SystemClassDB tenderly refactored 
SystemClassEDSM refactored all logic is in here to do the syncs
Controller refactored to order func in use, annotate them, simplify a little
Fixed crash in 3dmap due to not usin sorted list for distances
Rulset file added, will be turning on soon
Toolbar themeing reworked due to being applied to 3dmap.. Now actually uses toolbar themeing colours.  Themes have been improved some variants only shown if Font available
DraggableFormPos made to control saving/loading position - used by main form, 2d/3d/user control forms.   Implements temp resize functionality and knows not to save it.  Maximise now does not lose old size.


EDDiscovery/Forms/Form2DMap.cs
```diff
@@ -88,37 +75,44 @@ private void FormSagCarinaMission_Load(object sender, EventArgs e)
...
+            toolStripComboExpo.SelectedIndex = 0;
             toolStripComboBoxTime.SelectedIndex = 0;
             initdone = true;
             ShowSelectedImage();
+
+            imageViewer.BackColor = Color.FromArgb(5, 5, 5);
+
+            EDDiscovery.EDDTheme theme = EDDiscovery.EDDTheme.Instance;
+            bool winborder = theme.ApplyToForm(this);
+            statusStripCustom.Visible = panel_close.Visible = panel_minimize.Visible = !winborder;
```
phroggster: Probably a good idea to add a base class `protected ObservableList<Control> CaptionControls {get;}` or a `protected void Add/Remove-To/From-Caption(Control c) {}` to handle this stuff upstream.

EDDiscovery/Forms/GalaxySectorSelect.cs
```diff
@@ -0,0 +1,420 @@
...
+            Invalidate();
+        }
+
+        private void buttonExtSet_Click(object sender, EventArgs e)
+        {
+            List<int> currentsel = imageViewer.Selection;
+
+            System.Diagnostics.Debug.WriteLine("Initial Sectors " + string.Join(",", initiallist));
+            System.Diagnostics.Debug.WriteLine("Cur Sectors " + string.Join(",", currentsel));
+            bool added = false;
```
phroggster: `bool added = currentcell.Except(initiallist).Any();` to forgo the `foreach`.

EDDiscovery/UserControls/UserControlRoute.cs
```diff
@@ -110,7 +110,7 @@ private void ToggleButtons(bool state)
             int atpos = s.IndexOf('@');
             if (s != null && atpos != -1)
                 s = s.Substring(0, atpos);
-            s.Trim();
+            s = s.Trim();
```
phroggster: Needs a `null` coalesce/check? Just guessing based on the prior conditional.

EliteDangerous/DB/SystemClassDB.cs
```diff
@@ -199,11 +254,11 @@ public enum SystemAskType { AnyStars, PopulatedStars, UnPopulatedStars };
         }
 
 
-        public static List<long> GetEdsmIdsFromName(string name, SQLiteConnectionSystem cn = null , bool uselike = false)
+        public static List<long> GetEdsmIdsFromName(string name, SQLiteConnectionSystem cn = null, bool uselike = false)
         {
             List<long> ret = new List<long>();
 
-            if (name!=null && name.Length > 0)
+            if (name != null && name.Length > 0)
```
phroggster: `if (!string.IsNullOrEmpty(name))`

ExtendedControls/Forms/DraggableForm.cs
```diff
@@ -43,7 +43,7 @@ public DraggableForm()
         /// <seealso cref="OnCaptionMouseDown"/>
         /// <seealso cref="OnCaptionMouseUp"/>
         [DefaultValue(20), Description("The number of pixels from the top of the Form that will be interpreted as the Form's caption area.")]
```
phroggster: Need to update the default value attrib to match.

ExtendedControls/Themes/ThemeStandard.cs
```diff
@@ -949,6 +975,14 @@ public bool IsFontAvailableInTheme(string themename, out string fontwanted)
             return false;
         }
 
+        public bool IsFontAvailable(string fontwanted)
```
phroggster: Not a huge fan of this approach. Think that a full font replacement mapping, or the simplicity of falling back to MS Sans Serif, would be better. Now, can no longer select the fancy themes unless the fonts are installed, which means I'm not aware that other themes exist that want fonts, which means that I'm not looking for the fonts to download to try out the cool "new" theme.

EliteDangerous/DB/SystemClassDB.cs
phroggster: For the record, I ran a stopwatch on a few billion iterations of this a few weeks ago, comparing cost of null + Len checks vs. the static method over a wide variety of inputs. `string.IsNullOrEmpty` was roughly 75% of the cost overall, unless the test cases were all null (at which case it was about 102%).

EDDiscovery/Forms/GalaxySectorSelect.cs
phroggster: s/`currentcell`/`currentsel`/

PR #1662, opened 2018-01-26
MapPanel general fixes

andreaspada:
**General fixes and improvements**

- Fixed rotation (it still resets after zoom or pan; need refinement)
- Added pan (with the middle button, not available for null zoom)
- Added Zoom, centered under the mouse (almost)
- Stars are now 4 pixel, for better mouse hovering to display the tooltip
- Modified the textbox behavior to better handle the focus before to compute 

---

**Zoom**
The zoom is predefined to 1:1. Using the scroll wheel, it select through an array of additional ratios (up to 10:1). By now, when changing the zoom factor, it try to center the zoomed map to the cursor position.
A context menu is on the way, which let the user choose between "center on mouse" or "center on map" behavior. Also, it will be possible to directly choose a zoom ratio from the same context menu.

**Pan**
When the map is zoomed, it's possible to pan using the middle mouse button. I plan to let the user pan also with the keyboard arrows and/or wasd keys.

---

**Planned improvements**
- context menu which allow for change behavior of zoom (centered under mouse, centered on map middle)
- keyboard pan (arrows and/or WASD)
- context menu for the system under the mouse, for EDSM integration and history highlight


robbyxp1:
I know its here, just pending for a little ..

PR #1663, opened 2018-01-27
Fix Some Issues Noted in #1657

phroggster:
* Add `MouseUp` handlers to match `MouseDown` handlers for caption controls.
* Add copyright to `EDDiscovery/Forms/GalaxySectorSelect.cs`.
* Bump copyright year in `EDDiscovery/Forms/Form2DMap.cs`.
* Also sort the using directives in `EDDiscovery/Forms/Form2DMap.cs`.
* `DraggableFormPos`:
  * Default the `top` value to `int.MinValue` instead of `-999` to prevent collisions.
  * Use `this.DesignMode` instead of checking if options are instanced.
  * `Event` -> `OnEvent`, as it's cheaper and we can override/change behavior.
  * Don't save `Form` position if the close was cancelled.
    * Not sure why it was the way it was, but eventually it's going to get closed and that's the time to capture the info.
* Linq-ify some GalaxySectorSelect `foreach` blocks that weren't concerned about the items being iterated.
* Fix a couple NRE possibilities in `UserControlRoute`
* Use `SelectAll` in `UserControlRoute` instead of `Select(0,1000)` to select *all* of the text.
* Fix the `DefaultValue` attrib on `DraggableForm.CaptionHeight`, and add the default value to the XML doc for the property.

Not addressed in this PR but will handle separately:
* The revised theme handling code for missing fonts isn't ideal.
* `if (!string.IsNullOrEmpty(x))` is cheaper than `if (x != null && x.Length > 0)`, and easier to read.

### DIFF
--- a/EDDiscovery/3DMap/FormMap.Designer.cs
+++ b/EDDiscovery/3DMap/FormMap.Designer.cs
@@ -177,6 +177,7 @@ namespace EDDiscovery
             this.labelSystemCoords.TabIndex = 18;
             this.labelSystemCoords.Text = "Sol x=0.00";
             this.labelSystemCoords.MouseDown += new System.Windows.Forms.MouseEventHandler(this.panelTop_MouseDown);
+            this.labelSystemCoords.MouseUp += new System.Windows.Forms.MouseEventHandler(this.panelTop_MouseUp);
             // 
             // toolStripControls
             // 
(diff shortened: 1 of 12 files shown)


phroggster:
Yes, `DesignMode` fails under certain circumstances, but typically only when dealing with controls (most commonly due to inheritance). I've not seen Visual Studio fail to mark it appropriately on a `Form`, even when subclassed, but checking both conditions is an appropriate way to go.

I didn't peek too far into the GalaxySectorSelector changes, but they are representative of the code that was present, and at a fraction of the cost and complexity. I realize that we're not trying to write this in optimized x86 assembly language, but it's nice to use only what we need for the particular circumstances at hand.

PR #1664, opened 2018-01-27
Remove Lat/Lon from Location submitted to EDDN (9.0.x)

klightspeed:
Latitude and Longitude fields are disallowed in the EDDN journal schema, so any events that have these fields set (i.e. any undocked Location entries below about 2/3 of orbital cruise altitude) would have been rejected by EDDN.


robbyxp1:
I was not planning another 9.0.X, is this the end of the world?

klightspeed:
No - it just means that Location events where the commander starts undocked won't be accepted by EDDN.

AnthorNet:
I could remove them before next release, it's already planned to remove uploaderID to avoid combat logging and I'll check other potential personal information.

PR #1678, opened 2018-02-01
Map Panel general fixes and improvements.

andreaspada:
**General fixes and improvements**

- Fixed rotation (it still resets after zoom or pan; need refinement)
- Added pan (with the middle button, not available for null zoom)
- Added Zoom, centered under the mouse
- Stars are now 4 pixel, for better mouse hovering to display the tooltip
- Modified the textbox behavior to better handle the focus before to compute 
- Added context menu
- Added an About form, with usage instructions

---

**Zoom**
The zoom is predefined to 1:1. Using the scroll wheel, it select through an array of additional ratios (up to 10:1). By now, when changing the zoom factor, it try to center the zoomed map to the cursor position.
A context menu let the user to choose a zoom ratio, to center the map (mainly after pan) and to reset to the default position and size.

**Pan**
When the map is zoomed, it's possible to pan using the middle mouse button.

---

**Planned improvements**
- context menu option which allow for change behavior of zoom (centered under mouse, centered on map)
- keyboard pan (arrows and/or WASD)


andreaspada:
@robbyxp1 , I can't have the about form to be opened from the context menu. Can you kindly check what must be changed to have this properly done?

andreaspada:
Thank you so much, @klightspeed! I'll work on it immediately. KUDOS!

PR #1686, opened 2018-02-03
Don't feed EDSM back to EDDN

klightspeed:
* If only one cached system is found, only match it if the seached system has no coordinate
* Only override the system fields that should be overridden from the EDSM system

This should fix the bug that resulted in bad data being sent to EDDN.

This was reported on January 18 on the EDSM discord that multiple bodies were assigned to the wrong system.

> [10:50 AM] Irisa Nyira: I've visited and scanned both 2MASS J02324756+6127000 systems for the first time today, both in the same session. The celestial bodies I've scanned in one system seem to be appearing on both.
> [11:17 AM] RapidfireCRH: Do you have your journal entries for each star
> [11:17 AM] RapidfireCRH: that will help in splitting them up
> [12:11 PM] Irisa Nyira: Yes, I do. What should I do with it?
> [12:14 PM] RapidfireCRH: keep it handy if Anthor needs it
> [12:14 PM] Irisa Nyira: Will do
> [12:14 PM] RapidfireCRH: After i said that i thought he might just look at timestamps
> [12:14 PM] RapidfireCRH: but you never know
> [12:22 PM] Irisa Nyira: I'll keep them handy. If it helps, the 2MASS J02324756+6127000 at [-4267.281,96.781,-4228.125] is a binary system and both stars have planets, whereas the one at [-4232.500,96.000,-4193.656] is a quinternary system and no planets orbit A or B. There are a bunch orbiting ABCD, and some more orbiting E.

I then investigated the EDDB entries and reported in the Developers channel on the EDDiscovery discord

> [7:15 PM] Bravada Cadelanne: https://ross.eddb.io/eddn/log/6397337 it looks like the data from EDSM is overriding the position sometimes - the associated FSDJump is https://ross.eddb.io/eddn/log/6392198, while it looks like the sent Scan event took the system info from https://ross.eddb.io/eddn/log/6300094
> [7:16 PM] Bravada Cadelanne: Interestingly, a few minutes later, it took the correct system: https://ross.eddb.io/eddn/log/6399733
> [7:24 PM] Bravada Cadelanne: In fact, both systems are fairly old, so if it picked up the wrong system, either the system matching is faulty, or matching the body to the FSDJump didn't work properly.

### DIFF
--- a/EliteDangerous/DB/SystemCache.cs
+++ b/EliteDangerous/DB/SystemCache.cs
@@ -45,7 +45,7 @@ namespace EliteDangerousCore
             if (find.HasCoordinate && foundlist.Count > 0)           // if sys has a co-ord, find the best match within 0.5 ly
                 found = NearestTo(foundlist, find , 0.5);
 
-            if (found == null && foundlist.Count == 1)              // if we did not find one, but we have only 1 candidate, use it.
+            if (found == null && foundlist.Count == 1 && !find.HasCoordinate) // if we did not find one, but we have only 1 candidate, use it.
                 found = foundlist[0];
 
             if ( found == null )                                    // nope, no cache, so use the db
--- a/EliteDangerous/HistoryList/HistoryList.cs
+++ b/EliteDangerous/HistoryList/HistoryList.cs
@@ -493,15 +493,51 @@ namespace EliteDangerousCore
 
             if (edsmsys != null)
             {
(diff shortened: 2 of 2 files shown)


robbyxp1:
Only thing I'd say.. if we ever add another field to ISystem/Systemclass, are we going to remember to update that big manual copy constructor...

PR #1692, opened 2018-02-04
Add git version to AssemblyInformationalVersion

klightspeed:
This sets the AssemblyInformationalVersion to $AssemblyVersionStatic+$GitCommitHash

So `9.1.*` has the informational version `9.1+abc1234` where `abc1234` is the short commit hash.

If git isn't installed, or the build is performed from a sources tar.gz, the commit hash will be empty.


robbyxp1:
hmm see discord

PR #1710, opened 2018-02-08
Fix filename case in TestNetLogEntry Resources.resx

klightspeed:
Fixes f7389dedab9ced3bcaa46bb24f2e0a6b51f8472d


robbyxp1:
Bugger. Stupid case.  What is failing building it?

PR #1717, opened 2018-02-11
Adds option to Journal Grid export to export journal JSON

iainross:
Off the back of a conversation with CMDR Orvidius, allows a filtered view to be saved as a single journal file.  Can be uploaded to http://elite.necrobones.com/vid/travelvideo and possibly other end points in the future...


robbyxp1:
Should do travel history too... ;-) Will do it later.

Issue #1705, opened 2018-02-06
[9.0.4] Enhancement - Scan Data on System -> Move Position of Approx Value


Heavensburn:
Last one. :)

Screen space is expensive and with so many view that require width, I have a sidebar containing System Information, Scan Data on System and Nearest Stars.

However, I find with Scan Data on System, the Approx System Value still scrolls outside the viewable area. If possible, when you choose to Hide the Toolbar, could the Approx Value be moved to the left under the title (instead of disappearing). This is more meta data than a Toolbar option: 
![image](https://user-images.githubusercontent.com/13399258/35842215-93f53f9c-0ab5-11e8-97d4-37a8c9b9f0c8.png)

robbyxp1:
Been fixed.

PR #1724, opened 2018-02-14
Resolved 1705 Scan Data on System -> Move Position of Approx Value

Heavensburn:
Hi,

https://github.com/EDDiscovery/EDDiscovery/issues/1705

Now resolved.


robbyxp1:
I'm not sure what a flow panel brings to the party with WrapContents=false. Why did you use that? Why was it better than a standard panel.  I'm asking because we have had problems with the layout they bring previously, and all the other panels just use standard panels with fixed positions.. ;-)

robbyxp1:
Actually, a RollUpPanel may be a much better solution, then that bit can be hidden if needed..

Heavensburn:
Using Flow allowed me to collapse the horizontal such that:
image
Could be collapsed down to just the approx value:
image

But, I'll check out the RollUpPanel. Once I figure out how to get TestForm2 to compile and run. ;)

Heavensburn:
Hope that's better. 
On open:
2018-02-17 10_42_53-eddiscovery version 9 1 6622 19229
If you right click and choose to close the controlsPanel, 'Approx Value' will move over:
2018-02-17 10_42_10-eddiscovery version 9 1 6622 19229
If you unpin, everything will collapse:
2018-02-17 10_42_24-eddiscovery version 9 1 6622 19229

PR #1729, opened 2018-02-16
Catch and log DSERR_BUFFERLOST during Dispose

klightspeed:


robbyxp1:
I see great minds think alike.  that was the only thing i could think of doing

PR #1731, opened 2018-02-17
Fix some NOP `string.Trim()` calls, trim down a couple others

phroggster:
* Fix `ActionFile` and `SpeechSynthesizerWindows` that were using a NOP `string.Trim(...)` syntax.
* Rewrote a few other (repeated!) `string.Trim(...)` calls to use `string.IsNullOrWhiteSpace(...)`.


robbyxp1:
ta.  I hate that it runs but does not complain.  bring in later.

PR #1734, opened 2018-02-17
Revamp Audio library

phroggster:
* Bump `AudioExtensions` library from v1.1 to v1.2.
* Add platform-agnostic wrappers capable of choosing the best available engine for the designed task.
  * `AudioDriver`, `SpeechSynthesizer`, `VoiceRecognizer`.
  * Offloads the selection requirements to the `AudioExtensions` library.
  * However, caller can still specify a separate engine during construction. This is presently used when the audio is disabled.
  * Opens the door to ***safely*** add Mac, Linux, Android, whatever support, with few changes required, and simplifies construction.
* Moved `AudioDriver` ownership to `AudioQueue`, extended from `IDisposable`.
  * This greatly simplifies construction and destruction.
  * Further helps to isolate against `CSCore` dtor exceptions.
* `AudioExtensions.AudioQueue.AudioSample` -> `AudioExtensions.AudioSample`.
  * Track owning `AudioDriver` and automatically clean up after themselves.
  * Plugged leaks that were presumably causing the `CSCore` dtor exceptions.
* Much better implementations of `IDisposable` throughout.
* Removed all `System.Speech` references when `NO_SYSTEM_SPEECH` is defined.
* Set root namespace to match (the default namespace when adding a new file; was `Audio`, now `AudioExtensions`).
* `AudioExtensions.AudioQueue.Priority` -> `AudioExtensions.AudioQueuePriority`
  * Now defaulting to `Normal` if otherwise unspecified (was previously Low), yet retaining the existing sort order.
* Fix `ObjectDisposedException` if test sound finishes playing after a settings dialog gets disposed.
* Ignore theme if it isn't loaded.
  * Allows to add a simple debugging `TestAudioForm` jig that doesn't require setting up themes (not added, but easy to do so).


robbyxp1:
I'm not bringing this into 9.1. . its too big.  It will have to go to 9.2 and be tested for a while.

If there are minor bug crash fixes then push another PR with those in.

Cheers

rob

andreaspada:
Impressive, @phroggster!

robbyxp1:
Closing, this needs doing more gently, i.e a little at a time, so the changes are less per push and its more clear what is happening.  The risk is just too high.

PR #1735, opened 2018-02-17
Add BodyID processing to StarScan; process other events with body name/id

klightspeed:
This does not yet handle barycentres, as it is difficult to determine if a barycentre is a top-level barycentre or a planet-level barycentre, and what the names of those top-level barycentres are unless all top-level stars have been scanned.


robbyxp1:
I've already done it!  I did say in the chat.. i'll have to check how it clashes.

robbyxp1:
Sorry, comment is wrong, i've only done a small part.. i'll merge with that i've done..

PR #1737, opened 2018-02-19
Button for enable/disable the porocessing and display of jumponium materials

andreaspada:
Cleanup of the toolbar; added a button to show/hide jumponium materials, which also avoid to process their availability for a slight optimization, when not needed.


andreaspada:
Sorry, just noticed that StarDistance was modified without my knowledge. I'll close and clean those strange changes...

robbyxp1:
the strange changes fixed a race condition which crashed EDD regularly in 9.0.4!

andreaspada:
Ops...

PR #1739, opened 2018-02-19
Added a couple of switches to StarList

andreaspada:
Added a couple of switches: show/hide bodies classes and jumponium availability. It give both a clear view, if such information are not needed/requested and, if not needed, it pass over the nested condition to speed up things a bit (for example, for larges lists...)

### DIFF
--- a/EDDiscovery/UserControls/UserControlStarList.Designer.cs
+++ b/EDDiscovery/UserControls/UserControlStarList.Designer.cs
@@ -45,17 +45,18 @@ namespace EDDiscovery.UserControls
         {
             this.components = new System.ComponentModel.Container();
             System.ComponentModel.ComponentResourceManager resources = new System.ComponentModel.ComponentResourceManager(typeof(UserControlStarList));
+            System.Windows.Forms.DataGridViewCellStyle dataGridViewCellStyle1 = new System.Windows.Forms.DataGridViewCellStyle();
+            System.Windows.Forms.DataGridViewCellStyle dataGridViewCellStyle3 = new System.Windows.Forms.DataGridViewCellStyle();
+            System.Windows.Forms.DataGridViewCellStyle dataGridViewCellStyle4 = new System.Windows.Forms.DataGridViewCellStyle();
+            System.Windows.Forms.DataGridViewCellStyle dataGridViewCellStyle2 = new System.Windows.Forms.DataGridViewCellStyle();
             this.TopPanel = new System.Windows.Forms.Panel();
-            this.panel3 = new System.Windows.Forms.Panel();
             this.checkBoxMoveToTop = new ExtendedControls.CheckBoxCustom();
-            this.panel1 = new System.Windows.Forms.Panel();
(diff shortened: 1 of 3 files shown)


robbyxp1:
So, i don't mind the switches, but you've changed the formatting.. Travelgrid, Journal and Starlist should all share the same basic formatting of the control bar.. So you need to put it back to look the same as the other two.  The cursor to top should not be right docked as you end up with the damn thing over the more useful controls when you make the window smaller.  And icon sizes are 24/24 not 32/32.  Don't dock the buttons either, they should just be placed.

While your at it, remove the Star icon on starlist and move the time/search to the identical position as the journal/travelhistory.. ;-)  thats been missed.

robbyxp1:
I see you did remove the star icon.. ignore that part.

andreaspada:
Thanks for the review, @robbyxp1. That's exactly the kind of feedback I asked for: I'm new on many things, so sometime I don't clearly understand what can or cannot be changed. ;)

When you say:
> So, i don't mind the switches, 

It means that you do not see any advantage in having them? To me, they are quite necessary, as the information field can be sometime very large - I have systems with more than a dozen lines, between body classes and jumponium information. It can clutter a lot. Someone could prefer a simpler display, like the first implementation. Do you agree with that?

In regard the formatting:
> Travelgrid, Journal and Starlist should all share the same basic formatting of the control bar.. So you need to put it back to look the same as the other two

Of course! So, maybe, it should be better to implement the very same visual, and replace the switches with a "field filter" button? In that case, we can have a fine grained selection of information - say, **CMDR Potatos** want to be notified on every earth-like planet, but he do not mind for star classes. However, **CMDR Kremlin Strike** need to know everything about white dwarfs, neutron stars and black hole, and he also like Ammonia world... 
With the "Field Filter" button, we both have the very same visual of Travelgrid and Journal, more flexibility in regard the displayed information, and prepare in case we want more information to be displayed and filtered in or out - planet Eol Prou WY-R d4-282 2 has a planetary bookmark, i.e.

>  And icon sizes are 24/24 not 32/32

Ops! ;)

>  Don't dock the buttons either, they should just be placed.

Got it!

robbyxp1:
What i mean, is the extra buttons are fine.. just make them visually the same as the travelgrid/journal. Please don't change the look of the travelgrid/journal, make the starlist look like it but with the extra buttons.  And make sure the buttons look the same.

We have tried docking to the right before controls, but it looks awful when the window is made smaller, since the right controls override the left ones.  

And you don't need so many panels.. just one will do with the controls placed on it.

thanks.

andreaspada:
@robbyxp1, I reworked the panel as suggested by you. By now, I stuck with the switches.
I also implemented a SortCompare override and allow Visits column to be treated as integer, to try to fix #1741, but it do not work. The column data continue to be sorted as strings. Need a look by an expert.

robbyxp1:
This will have to be post 9.1, 9.1.1 etc. I don't have the bandwidth over the next few days to do anything on this.

PR #1743, opened 2018-02-21
Added one more panel position to History.

andreaspada:
Splitted the bottom panel - under history - in two, to permit a better distribution of sub-panels.

### DIFF
--- a/EDDiscovery/UserControls/UserControlCommonBase.cs
+++ b/EDDiscovery/UserControls/UserControlCommonBase.cs
@@ -34,6 +34,7 @@ namespace EDDiscovery.UserControls
         protected const int DisplayNumberHistoryBotRight = 1001;
         protected const int DisplayNumberHistoryMidRight = 1002;
         protected const int DisplayNumberHistoryTopRight = 1003;
+        protected const int DisplayNumberHistoryBotCenter = 1004;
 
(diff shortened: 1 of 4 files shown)


robbyxp1:
I'm going to have to consider this,.. the History UI is already cluttered.. Now your asking for another panel.

If they want multiple panels, they can do that via the grid.

Its rejected for now, if we must do this, we should make the whole thing configurable..

andreaspada:
No problem. Indeed, I think that history should be another usercontroller, contained in a grid.

andreaspada:
@robbyxp1, now that tabs can be renamed, Grid is actually much more usable than History tab. However, History has a very interesting and useful behavior, that Grid do not have: automatic resize of contained control when drag their borders, as they are docked inside split panels.
I propose some addons to Grid, to fully replace the current implementation: 

- behavior selection (manual or automatic tiling). 
- presets combobox, with same defaults combinations (one of them which mimics the current default History tab) and newly user-created presets.
- the controls should not be able to overlap.
- when in automatic tiling, the grid should divide itself in halves (or another user defined ratio), like a simple implementation of DWM. When a UserControl is created, the grid should open a space and create that control in it.
- each UserControl should have a position-id and a memberofgrid-id, so they could be programmatically moved in order and from a grid to another.

PR #1746, opened 2018-02-22
Added a couple of switches to StarList panel; fix for #1741

### LINKED ISSUE #1741: Sorting in Star list by visits results in incorrect ordering
If you click to sort the Star list by visits, it has issues with double digit visit counts, as seen in the attached screenshot.
screenshot 3878

andreaspada:
- Show/Hide Body classes and Jumponium notifications fora cleaner list when needed.
- Fixed #1741 (I bypassed the DataGridViewSorter call for the Visits column, as it was the offending code.)
- Uniformed the toolbar as History and Journal panels

### DIFF
--- a/EDDiscovery/UserControls/UserControlStarList.Designer.cs
+++ b/EDDiscovery/UserControls/UserControlStarList.Designer.cs
@@ -44,76 +44,55 @@ namespace EDDiscovery.UserControls
         private void InitializeComponent()
         {
             this.components = new System.ComponentModel.Container();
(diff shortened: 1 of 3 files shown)


robbyxp1:
PR1746 is in, but the buttons are not right.. You should be using CheckBoxCustom..  you should not be using local resource based bitmaps ever.  You have the save button stuff wrongly named in the db strings.  I have now fixed it on the master.

before you add new icons, you need to understand how we are managing icons now in EDDIconSet ..

PR #1814, opened 2018-03-04
Handle truncated api-v1/systems response

klightspeed:


robbyxp1:
So, given that EDSM was underload yesterday (prob. due to everyone upgrading to 3.0 tool sets), is this why you did this?  And is it relevant for the future?

robbyxp1:
Ie. do we expect this to be a permanent problem..

klightspeed:
It looks like @AnthorNet did something to the PHP config to kill PHP requests that take too much CPU or memory to process.  Requesting 3 hours of data instead of the current 12 hours of data currently does not run foul of the request killer.

We would probably need to avoid doing an update sync if the data is more than say 30 days old (by which point the full sync should kick in), and reduce the requested period from 12 hours to e.g. 3 hours.

A `LIMIT 10000` or similar on EDSM's side might also be in order to avoid running into the request killer (12 hours of data is about 30000 entries)

The request killer killing requests also affects `api-v1/sphere-systems` and `api-v1/cube-systems` - probably need to limit the radius of those requests to e.g. 100ly within about 2kly of Sol / Sag A* / Colonia.

AnthorNet:
I haven't changed those API for a couple of month, do you have any specific example so I take a look tomorrow?

robbyxp1:
Let co-ord with Anthor before we bring in these two.. he has been being attacked this weekend and although these may solve it they may not be needed long term...

klightspeed:
When requesting 3 hours or more of data using gzip compression, or 4 hours or more of data without gzip compression, the SSL connection is sometimes killed after sending 5 x 16KiB PDUs + 5 packets from a 6th incomplete PDU.

https://drive.google.com/drive/folders/1lL4ZAY59MdJcjSaV4j41POIq5BYd3MAT

robbyxp1:
Lets hold this in our back pocket for now and see how EDSM reacts in the future.

robbyxp1:
Retiring this

PR #1816, opened 2018-03-04
Get systems in 3 hour blocks; limit update to 30 days

klightspeed:
Requesting more than 3 hours causes EDSM to kill the request after a random number of systems have been retrieved.


robbyxp1:
Anthor is going to have a think and come back to us tomorrow on his thoughts - lets hold before we change anything.

robbyxp1:
See discord. Anthor does not want this.

PR #1819, opened 2018-03-04
Added Carbon in the list of jumponium materials.

andreaspada:


iainross:
It's probably worth trying to leverage Recipes.SynthesisRecipes for this rather than having a separate hard-wired list of ingredients that needs to be maintained independently.

You also don't seem to have added Ge to the Premium boost ingredients - that needs to go in as well as C (and will happen automatically if you drive this off the static Recipes class)

andreaspada:
Yes, @iainross. I'll give it a look. The less we repeat, the better and safer. ;)
Thanks a lot!

robbyxp1:
As Iain says, need rework. Resubmit later.

PR #1838, opened 2018-03-10
Fix `UserControlExpedition` `SQLiteException`

### LINKED ISSUE #1837: SQLiteException after Save Route
### Description
I deleted one of my routes. Imported a new route from an exported EDSM file. And named the route the same as the one I deleted. After clicking on **Save Route** I got the following error :

phroggster:
* `SavedRouteClass.UpdateDBFromExpeditionFiles`:
  * Fix several `InvalidOperationException`s if no `Start`/`End` `DateTime?` values are set.
  * Deleted routes now stay deleted even if they get updated upstream.
  * Deleting a route now sets the `changed` flag for consumers to see.
* `UserControlExpedition`:
  * Delete custom routes added by the user when clicking delete.
  * Only mark routes as faux-"deleted" if they come from EDSM.
  * Fix #1837

### DIFF
--- a/EDDiscovery/UserControls/UserControlExpedition.cs
+++ b/EDDiscovery/UserControls/UserControlExpedition.cs
@@ -13,7 +13,6 @@
  *
  * EDDiscovery is not affiliated with Frontier Developments plc.
  */
-using EDDiscovery.UserControls;
 using EliteDangerousCore;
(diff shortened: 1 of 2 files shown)


robbyxp1:
Okay, will bring into the release this week for the next ED version. thanks

PR #1842, opened 2018-03-12
Statistics from journal added to new tab in Stats UC

iainross:
Stats UC now remembers what tab was selected on reload
TreeViewCustom available in ExtendedControls if a themeable, scrollable TreeView is needed again elsewhere.


andreaspada:
Awesome! Thank you so much, @iainross!

PR #1854, opened 2018-03-20
Fix inverted json match in FindEntry

### LINKED ISSUE #1853: Two events of type DiscoveryScan with same time stamp, only first shows in History
Upon landing in a system, I get an event with type "DiscoveryScan" for "discovering" the star, and another event once my ADS fully charges.  Normally, I see both events if they happen in different seconds.  However, if they happen in the same second, I only see the first event.

Please see events below copy/paste from journal and screenshot of History grid, and please let me know if you need more information.

{ "timestamp":"2018-03-20T02:17:24Z", "event":"DiscoveryScan", "SystemAddress":354326415971, "Bodies":1 }
{ "timestamp":"2018-03-20T02:17:24Z", "event":"DiscoveryScan", "SystemAddress":354326415971, "Bodies":13 }

discoveryscan

klightspeed:
This should fix #1853, at least for `DiscoveryScan` events with different numbers.

I don't know if it's worth trying to handle the edge case of two subsequent `DiscoveryScan` events coming in with the same number of bodies in the same second but in separate reader ticks.  Would probably have to group events by timestamp and keep the latest events read in the last tick around to be handled in the next tick.  This would add up to 2 seconds extra delay in recognizing events given a 2 second tick.  This would also require rewinding the reader to the start of the first event in that second in order to prevent missing events.

### DIFF
--- a/EliteDangerous/EliteDangerous/JournalEntry.cs
+++ b/EliteDangerous/EliteDangerous/JournalEntry.cs
@@ -971,7 +971,7 @@ namespace EliteDangerousCore
                         while (reader.Read())
                         {
                             JournalEntry jent = CreateJournalEntry(reader);
-                            if (!AreSameEntry(ent, jent, entjo))
+                            if (AreSameEntry(ent, jent, entjo))
                             {
                                 entries.Add(jent);
                             }


robbyxp1:
Bit confused..
Monitor watcher only wants to add entries which are not already in the DB
ents = ents.Where(jre => JournalEntry.FindEntry(jre.JournalEntry, jre.Json).Count == 0).ToList();

So you don't want to find another one, but the current code is giving you the list of ones which are not the same entry, (!Are..) so I agree.

Don't do what your saying above. We don't want events delayed for any reasons, it would ruin the voice system and responsiveness ;-)

And this is a corner case.

PR #1860, opened 2018-03-24
In-game stat tree stat saved reliably, Home System validation no longer prevents entry

iainross:


robbyxp1:
Using leave is problamatic normally.. its not guaranteed someone will leave the control.  Its normally better to validate as they type, using text changed..

robbyxp1:
I've removed quite a lot of leave validations for that reason..

iainross:
Aye - that's why it does a check on closing as well.  The problem with Validated is that it was over-writing with the setting as you typed making it impossible to actually change it.

robbyxp1:
i'll have a think if we can improvie it after i merge it in

PR #1874, opened 2018-03-28
Plot addons

andreaspada:
- Added a checkbox to enable/disable the virtual depth axis - which controls the dot size
- Previous system is colored in purple, for a better visualization of the exploration path
- Included an icon for the checkbox


andreaspada:
@robbyxp1, I reworked and cleaned a bit the PR. Now, dots.png is where it is supposed to be, and finally I understood how to successfully include and use new icons.

andreaspada:
@robbyxp1, I need some help to solve the tooltip of Plot. They are implemented with the same logic of LocalMap (which is working as expected in that regard) however, the un-visited systems ALL share the same name in the tooltip. It does not appear to happens in visited systems, although, which are named as expected, each one with its name in the tooltip.
Can you kindly check to see if yo notice how to solve that? It's quite important to check if, finally, the coordinates are oriented as we need. ;)

PR #1875, opened 2018-03-28
Added nzy3d-api

andreaspada:
I included the nzy3d-api - renamed Charts, for a cleaner folder tree - which successfully build. It can be used as main replacement for plots, charts and maps, accelerated 3D and 2D.

I'm currently working porting the LocalMap and Plot to this api, to test its performance and usability.


klightspeed:
It looks like fixing those references actually broke the build.  You're pointing at the .NET 2.0 OpenTK assembly when you should be pointing at the .NET 4.0 or .NET 4.5 assembly

klightspeed:
Correction: AxeBox.cs references `OpenTK.Graphics.RenderingMode`, which doesn't exist in recent versions of OpenTK.

PR #1884, opened 2018-03-31
Tidied up journal export to remove excess formatting

iainross:


robbyxp1:
To be honest, this is a bit like putting a quart into a pint pot.  CVSWriteGrid is designed to write CVS files, as as such, they need the Format() call to ensure they are properly formatted on output. 

Your trying to use this writer as a text writer not as a CVS writer.. 

are you sure it would not be better just to take away the writing of journals from it at all and write new code in UCJG.cs.

PR #1885, opened 2018-03-31
Include Location in messages sent to EDDN

klightspeed:
EDDN Sync was set to send Location events, but they were not included in the types of events to send to EDDN.


robbyxp1:
Cheers.. not sure ISEDDNMessage should really be in HistoryEntry - its not really its job to know, more like EDDNClass.cs..
minor point.

PR #1895, opened 2018-04-02
Sgr A* Value now reported correctly

iainross:


AnthorNet:
Does it means SuperMassive is lowest at sale than regular?

iainross:
Yes, we were previously estimating Sgr A* at ~23million rather than the 623k you actually get - and that was with the default star kValue, for the regular black hole one it would be worth 423 million!!!

AnthorNet:
Thanks! Fixing on my side too ^^

PR #1904, opened 2018-04-09
Rewrite of PlotPanel. 

andreaspada:
- Logic and code cleanup. 
- Switch to OxyPlot, from NuGet, for better performance and overall improvements.

**Details**
The Plot uses real coordinates, now, and not relative computed distances. That position the systems in their real life position relative to each other, from each and every POV. There's also a new path which can be setup in the Settings panel, which is used to export plots views in PNG. 

User interaction: 
- Pan* | Right mouse button
- Zoom* | Mouse wheel
- Zoom by rectangle* | Ctrl+Right mouse button, Middle mouse button
- Reset* | Ctrl+Right mouse button double-click, Middle mouse button double-click
- Show ‘tracker’ | Left mouse button


robbyxp1:
I'm going to reject this.

Why are you altering lots of packages which are not part of this pull.  Why am i seeing version changes to JSON, directinput, Directinput, etc.  I don't want to change them at this point.

Settings is also wrong.. it should not be doing SQLite stuff directly, it should be going thru EDDConfig.  Why are you editing the screen shots folder? its already in the sub dialog.

TestExtendedControls is also about testing stuff in the extendedcontrols package..  Its not the place to check oxyplot.

You need to give me a push affecting only the plot panel and a few other files to get your new DLL package  in.

andreaspada:
Ok.

PR #1921, opened 2018-04-29
Materials UC now shows material limits as well as counts

iainross:
Fixed a couple of incorrect types as well

### DIFF
--- a/EDDiscovery/UserControls/UserControlMaterialCommodities.cs
+++ b/EDDiscovery/UserControls/UserControlMaterialCommodities.cs
@@ -117,7 +117,7 @@ namespace EDDiscovery.UserControls
 
                     if (materials)
                     {
-                        rowobj = new[] { m.name, m.shortname, m.category, m.type, m.count.ToString() };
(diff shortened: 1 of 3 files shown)


EDDiscovery/UserControls/UserControlMaterialCommodities.cs
```diff
@@ -117,7 +117,7 @@ private void Display(MaterialCommoditiesList mcl)
 
                     if (materials)
                     {
-                        rowobj = new[] { m.name, m.shortname, m.category, m.type, m.count.ToString() };
+                        rowobj = new[] { m.name, m.shortname, m.category, m.type, $"{m.count.ToString()}/{MaterialCommodityDB.MaterialLimit(m.type)}" };
```
robbyxp1: Can't this return null?

robbyxp1:
Can you check my point above and see if i'm right before i pull this in..

robbyxp1:
going to merge anyway

iainross:
Probably can but it won't error just show x/ rather than x/y.

Get Outlook for Android<https://aka.ms/ghei36>

________________________________

PR #1928, opened 2018-05-05
Shopping lists in tabs no longer error on startup

iainross:


robbyxp1:
Yep, this was a contentious area.. I had to change the interface behaviour at start up and where the uctg was set up changed..

uctg is only null during the Init call..  Can't see a path in synth or engineering to use it before init finishes...

I'm worried its showing up a behaviour the framework is not doing right... I mean you must have a UCTG at the load layout point. Is there a path where the cursor UCTG is not set up right?

I've put the SL in a tab and it does not crash on mine (without this change).

robbyxp1:
With it crashing, can you see if you have uctg != null at the LoadLayout, and show me the path where  SetHistoric is called during init and before the uctg is set..

iainross:
UserControlEngineering.Init sets the checkbox on line 119 which goes, via the _CheckedChanged event, into SetHistoric.  That is called from the ShoppingList's .Init

Gets in there reliably for me with an historic shopping list in a tab during startup (not for one added after it's running).

Moving the checkbox setting into LoadLayout does appear to also fix it.

iainross:
And is probably better, updated commit will be along shortly.

robbyxp1:
Cheers!

PR #1940, opened 2018-05-12
Fix NRE when history cursor is missing

### LINKED ISSUE #1939: error prompt when opening 3d map

klightspeed:
This should fix #1939


robbyxp1:
So, lets see what the user says first.  Its obv not a user wide issue - only 1 report. And as I said in discord, we should never need to cope with the Primary Tab/TG being missing - by design it must be there, and its set up in CreateTabs (then CreateTab then Init in UCCB which creates the sub pages) so by the end of Discovery form init it must be there.

robbyxp1:
Going to not incorporate this - have put extra checked in instead

PR #1941, opened 2018-05-12
SPanel now includes current material levels in scan info

iainross:
Journal scan display string no longer shows 0/# for all mats when no material list is passed in.


robbyxp1:
Ooops.. forgot you may not want them.  Thanks

PR #1943, opened 2018-05-13
 Fix queued events being held too long

klightspeed:
This bug is resulting in any events read before `LoadGame` being placed after the last entry read.

Fixes: 99850a0206735f2a8087d556345ac9048f7a2376

### DIFF
--- a/EliteDangerous/EliteDangerous/EDJournalReader.cs
+++ b/EliteDangerous/EliteDangerous/EDJournalReader.cs
@@ -211,14 +211,6 @@ namespace EliteDangerousCore
         {
             jent = new List<JournalReaderEntry>();
 
-            if (StartEntries.Count != 0 && this.TravelLogUnit.CommanderId != null && this.TravelLogUnit.CommanderId >= 0)
-            {
-                jent.Add(StartEntries.Dequeue());
-                //System.Diagnostics.Debug.WriteLine("*** UnDelay " + jent[0].JournalEntry.EventTypeStr);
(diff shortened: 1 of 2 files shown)


robbyxp1:
So i agree with moving it inside the loop, but should it not be at the top of the loop..

            jent = new List<JournalReaderEntry>();

            bool readanything = false;

            while (ReadLine(out JournalReaderEntry newentry, l => ProcessLine(l, resetOnError)))
            {
                readanything = true;

                if (newentry != null)                           // if we got a record back, we may not because it may not be valid or be rejected..
                {
                    if ((this.TravelLogUnit.CommanderId == null || this.TravelLogUnit.CommanderId < 0) && newentry.JournalEntry.EventTypeID != JournalTypeEnum.LoadGame)
                    {
                        //System.Diagnostics.Debug.WriteLine("*** Delay " + newentry.JournalEntry.EventTypeStr);
                        StartEntries.Enqueue(newentry);         // queue..
                    }
                    else
                    {
                        while (StartEntries.Count != 0)
                        {
                            var dentry = StartEntries.Dequeue();
                            dentry.JournalEntry.CommanderId = (int)TravelLogUnit.CommanderId;
                            //System.Diagnostics.Debug.WriteLine("*** UnDelay " + jent[0].JournalEntry.EventTypeStr);
                            jent.Add(dentry);
                        }

                        //System.Diagnostics.Debug.WriteLine("*** Send  " + newentry.JournalEntry.EventTypeStr);
                        jent.Add(newentry);                     // else pass back
                    }
                }
            }

            return readanything;
        }

PR #1956, opened 2018-05-16
Fix typo causing travel counters not to reset

### LINKED ISSUE #1954: Trip computer
Once again the question about the reset of the travel distance in the trip computer when the Start button is pressed.
(I refer to the ticket #1942)

If I have understood correctly, a starting point should be set in the journal when the Start button is pressed and the travel distance in the trip computer will be reset at the next journal entry.
(I may not have understood it correctly. My English is not good and I have to work with a translation program. And what comes out sometimes, you sure know yourself too well.)

However, the problem with the reset is still there. The starting point in the journal is set, but the travel distance in the trip computer is not reset.

I don't know if it helps anything, but appendix are the log files since my last ticket.
[Log.zip](https://github.com/EDDiscovery/EDDiscovery/files/2009006/Log.zip)

Thank you for your efforts

klightspeed:
This should fix #1954

### DIFF
--- a/EDDiscovery/UserControls/UserControlTrippanel.cs
+++ b/EDDiscovery/UserControls/UserControlTrippanel.cs
@@ -280,7 +280,7 @@ namespace EDDiscovery.UserControls
                     if (list.Count() > 1 && he.isTravelling)
                     {
                         he = list.ToArray()[1];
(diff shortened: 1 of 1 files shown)


robbyxp1:
Thanks!

PR #1992, opened 2018-06-13
Convert EventSummaryName to a lazy property

### LINKED ISSUE #1987: Unhandled Exception

### LINKED ISSUE #1991: Unhandled UI Exception

klightspeed:
This should fix #1987


robbyxp1:
Cheers, but.. isn't is just simpler, in the only other place where JournalEntry can be made outside of the JSON creation chain (which does fill it it), to do:

   public JournalEntry(DateTime utc, int synced , JournalTypeEnum jtype)       // manual creation via NEW
        {
            EventTypeID = jtype;
            EventTimeUTC = utc;
            Synced = synced;
            TLUId = 0;
            EventSummaryName = FillSummary;     // after creation, so journal fields are populated.
        }

I think thats cleaner and less lines changed etc..  I'll commit this change for now.

PR #1993, opened 2018-06-13
Convert EventSummaryName to a lazy property (v10.0.x)

klightspeed:


robbyxp1:
Hold not sure another 10.0.X is going to be done

Issue #1994, opened 2018-06-14
Unhandled UI Exception+Inara send problems


NecroBones:
Right after starting up with 10.1, and adding Inara info, and then clicked the button to send historic Inara data:

### Exception Details:
>EDDiscovery v10.1.0.0 There was an unhandled UI exception.
>```
>==== BEGIN ====
>System.NullReferenceException: Object reference not set to an instance of an object.
>   at EliteDangerousCore.Inara.InaraSync.HistoricList(HistoryList history) in C:\Code\EDDiscovery\EliteDangerous\Inara\InaraSync.cs:line 114
>   at EliteDangerousCore.Inara.InaraSync.HistoricData(Action`1 logger, HistoryList history, EDCommander cmdr) in C:\Code\EDDiscovery\EliteDangerous\Inara\InaraSync.cs:line 57
>   at EDDiscovery.EDDiscoveryForm.sendHistoricDataToInaraToolStripMenuItem_Click(Object sender, EventArgs e) in C:\Code\EDDiscovery\EDDiscovery\EDDiscoveryForm.cs:line 1187
>   at System.Windows.Forms.ToolStripItem.RaiseEvent(Object key, EventArgs e)
>   at System.Windows.Forms.ToolStripMenuItem.OnClick(EventArgs e)
>   at System.Windows.Forms.ToolStripItem.HandleClick(EventArgs e)
>   at System.Windows.Forms.ToolStripItem.HandleMouseUp(MouseEventArgs e)
>   at System.Windows.Forms.ToolStrip.OnMouseUp(MouseEventArgs mea)
>   at System.Windows.Forms.ToolStripDropDown.OnMouseUp(MouseEventArgs mea)
>   at System.Windows.Forms.Control.WmMouseUp(Message& m, MouseButtons button, Int32 clicks)
>   at System.Windows.Forms.Control.WndProc(Message& m)
>   at System.Windows.Forms.ToolStrip.WndProc(Message& m)
>   at System.Windows.Forms.ToolStripDropDown.WndProc(Message& m)
>   at System.Windows.Forms.NativeWindow.Callback(IntPtr hWnd, Int32 msg, IntPtr wparam, IntPtr lparam)
>===== END =====
>```

robbyxp1:
You have no ships stored.. okay i'll protect against that.  At least its in the historic upload.

robbyxp1:
Does it always happen? did it happen during loading?

NecroBones:
[...]

PR #1996, opened 2018-06-15
Master shipfd always

klightspeed:
This should help fix the SRV overload and missing ship names in #1994


robbyxp1:
I'll have a look, but i'm in the middle of tracking it down as well with user logs.. hang on a sec.

PR #1998, opened 2018-06-15
Ensure ShipFD is always set

klightspeed:


robbyxp1:
Please wait.. i'm in the middle of tracking it down.. and I want the SRV/Fighter in there.

robbyxp1:
So, we have clashed, and i've found a fundamental bug with the module buy.. so this is not needed. thanks

PR #2017, opened 2018-06-18
Fill summary in EDDCommodityPrices

klightspeed:
Fix another place where `EventSummaryName` was not set.
This should hopefully plug #2016 (though the proper fix would be #1992)


robbyxp1:
Cheers.  The only thing about the lazy load is that it made make the history take longer to load.. not sure one way or another, but i did a lot of testing with the current implementation both with EventSummary computed as it went and with it doing it as now.  Hence my sort of reluctance to change the method

klightspeed:
#1992 should have only slowed it down in cases where it would currently throw an NRE, as the pre-population of the summary was still done.

PR #2035, opened 2018-06-25
Added some predefined size ratios to spliiter panels

spadino:


robbyxp1:
I'll have to look at this

PR #2039, opened 2018-06-28
Fix nomenclature

spadino:
Changed "Stars" to "Systems", for a more coherent and astrophysics true nomenclature. Till now, stars and systems name are used in a somewhat confusing manner: we shouldn't refer to a "star systems" (coherent, gravitational bounded or isolated masses) as "stars", as they can be formed by planets alone (rouge planets...).


robbyxp1:
Can you wait until i've done the basic translation architecture before we do this.. it will change stuff like this.

And we have to stop renaming classes, since they are ref in the tx code.

PR #2048, opened 2018-06-29
Prevent Index out of range errors in key parser

### LINKED ISSUE #2038: was working fine for a few days then all the sudden it started crashing on load and i cant use it

klightspeed:
This should fix #2038


robbyxp1:
Ta. 

I'm trying my hardest to see where this could come from, considering the keys are all defined in the free voice control pack and should not be user/game changeable..  is there an error in the pack somewhere?

robbyxp1:
Got it now. cheers

robbyxp1:
Applied to 10.1.x also

PR #2065, opened 2018-07-06
Make UI responsive when loading grids

klightspeed:
These use windows timers to process the creation of rows in chunks.  As WM_TIMER events are only generated when no other events are in the queue (see https://blogs.msdn.microsoft.com/oldnewthing/20111219-00/?p=8863/), this will allow the UI to process events between the processing of grid chunks.

Real-time journal events are queued while the grid is loading, and dispatched once the grid is loaded.

In order to reduce the time before anything is displayed, rows are added in chunks to the grids rather than being kept in a collection and added all at once after the are all created.  Note that row creation is the slowest part of the grid population - it takes about 20s for 40k rows to be created and filled, while it takes a couple of hundred ms to put those rows into the grid and filter them.

### DIFF
--- a/EDDiscovery/UserControls/UserControlJournalGrid.cs
+++ b/EDDiscovery/UserControls/UserControlJournalGrid.cs
@@ -66,6 +66,10 @@ namespace EDDiscovery.UserControls
 
         Timer searchtimer;
 
+        Timer todotimer;
+        Queue<Action> todo = new Queue<Action>();
+        bool loadcomplete;
+
         public UserControlJournalGrid()
         {
             InitializeComponent();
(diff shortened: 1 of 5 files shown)


robbyxp1:
Is this synced to the current master? I'm splatting all over the code at the moment due to the translation system.. hate to lose stuff.

klightspeed:
Yes, it is synced to master.

robbyxp1:
Just trying it, nice idea.. But

On 10.1.5, with a 420k log entry file, TG load takes 16.1 seconds to fill out every entry.

On your PR, it takes 51s.

I know you get the top first, but the UI is still showing the wait cursor, the other panels are not being painted until the TG completes, the wait cursor is still on.

I think it needs more work, but i like it.  I wonder why it's taking so much more time?

klightspeed:
I have rebased my work onto current master, and have switched to loading the grids in one shot, which should reduce UI paint delay.

robbyxp1:
Thanks. I think it can be optimised further.. 

I removed the older IHistoryCursor callback with row/col (not used) and kept the one with the he/hl.

Now, we could fire off the FireChangedSelection before the end of painting, by working out which entry we are going to put the cursor on.  This could be done immediately before painting and thus make the other panels redraw quicker...

PR #2070, opened 2018-07-12
ScanData dirty hack.

spadino:
ScanData now disappear along with the main content if entered System/Galaxy map or Docked, if preferred.


spadino:
Closing. Better solution coming...

PR #2082, opened 2018-07-23
Added preliminary Italian translation

spadino:


robbyxp1:
Not the right name for it.. it needs to be Italiano-it.tlf (two letter ISO name)

robbyxp1:
see my comment in translate-eddiscovery as well about the <>

spadino:
Closing, and updating name and removing <> to a new branch.

PR #2093, opened 2018-07-27
Updates to italiano-it.tlf and some works on portugues-pt.tlf

spadino:
Italian translation file to lower case, as the others. Some additional translations.


robbyxp1:
there are coding errors in here

PR #2096, opened 2018-07-28
Fix dotless i (\u0131)

### LINKED ISSUE #2095: Unicode Character 'LATIN SMALL LETTER DOTLESS I' (U+0131)
EDDiscovery v10.1.6.0 is using U+0131 for some commodity names.
see: https://ross.eddb.io/eddn/log/76905490

```
{
  "\u0131nsulatingmembrane",
  "neofabric\u0131nsulation",
  "heatsink\u0131nterlink",
  "\u0131ndium",
  "\u0131ndite"
}
```

klightspeed:
This should fix #2095

### DIFF
--- a/BaseUtils/Files/CSVRead.cs
+++ b/BaseUtils/Files/CSVRead.cs
@@ -107,7 +107,7 @@ namespace BaseUtils
 
             static public int CellNameToIndex(string s)
             {
-                s = s.ToLower();
+                s = s.ToLowerInvariant();
                 int col = int.MaxValue;
                 if (s.Length >= 1)      // later, cope with An, Bn etc.
                     col = s[0] - 'a';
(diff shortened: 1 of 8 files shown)


robbyxp1:
Need to understand this more..

EDDNClass.cs sends fdname (389) to EDDN

        "data":"ındium",
        "message":"unknown commodity name"

where is this coming from?  Indium in materialcommds is okay - its an i.  fdname should not be localised?

klightspeed:
It looks like Frontier is using `ToLower` to lowercase the commodity symbolic name.

hchalkley:
I don't understand how the dotless i would get into the market.json, there's no tolower() applied to the string written out, and anyway Frontier's Cobra engine fToLower() method doesn't check locale, and only converts within the plain ascii range.
I have logged the issue as ED-99647 in our Jira database, but further evidence (such as a player journal or market.json fie) would help

klightspeed:
Gah - it isn't Market.json, but rather https://github.com/EDDiscovery/EDDiscovery/blob/master/EliteDangerous/EliteDangerous/CommodityPrices.cs#L74

PR #2101, opened 2018-07-29
Don't leave the body until LeaveBody

klightspeed:
This should fix incorrect distances shown in the Compass while in orbital cruise, after entering supercruise.

### DIFF
--- a/EliteDangerous/HistoryList/HistoryEntry.cs
+++ b/EliteDangerous/HistoryList/HistoryEntry.cs
@@ -42,10 +42,6 @@ namespace EliteDangerousCore
         public string Group { get; private set; } = "";                   // group..
         public bool? Wanted { get; private set; } = null;
         public string StarSystem { get; private set; }
-        public long? SystemAddress { get; private set; }
-        public float StarPosX { get; private set; } = Single.NaN;
-        public float StarPosY { get; private set; } = Single.NaN;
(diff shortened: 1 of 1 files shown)


robbyxp1:
Looks good, why do we need to fix it up on undocking?  Is that just for completeness?

klightspeed:
`ApproachBody` is not logged when the commander logs in docked to a planetary outpost, nor is it logged when we undock from said outpost.

robbyxp1:
Thanks.  I was wondering more about the dictionary.. which makes the class have a memory, where it did not have one before.  Which is not ideal.

So I get that we want to restore the state we had before docking (whereami/bodytype).  I've been thinking of a way to do it without memory but i can't see one either.

Would not just a record of these two parameters, written on dock, be enough to restore it?  Could that dictionary get large on very large data sets (being memory paranoid)? Do we need the dictionary of market ids to guarantee restore?..

Hmmm..

Anyway i'll incorporate this in for now. thanks.

robbyxp1:
Some comments..

HES constructor not copying StarPos?  StarPos? is not in use.
Constructor not copying SystemAddress..  SystemAddress is not in use

Need a little tlc ;-)

robbyxp1:
I've taken a branch and fixed up a few issue i saw when it processed my history on this new branch.

I've also simplified a few things, such as having an enum for travel state, and removing some fields..

will check and post a new PR

PR #2111, opened 2018-08-05
Some corrections to italiano UTF-8

spadino:


robbyxp1:
Hmm
this will delete portugues from the master..

spadino:
Yes, @robbyxp1, is intended. I have the Portugues on another branch. It's too early to have it, by now. Too many UTF-8 errors, and lacks many things. I'll push another one today or tomorrow later...

PR #2113, opened 2018-08-07
Make an OpenTKUtils project and move relevant code to it

robbyxp1:
Also split apart the data3d stuff and rename to make it clearer the relationship

OpenTK versions maintained.  No code changes


spadino:
So, if someone want to write another opentk form or panel - LocalMap, I'm thinking on you... - OpenTKUtils can be re-used, that's right?

robbyxp1:
Yes, but dont yet ;-)

PR #2114, opened 2018-08-07
Portugues

spadino:


spadino:
@robbyxp1, it think there's an Italiano-it.tlf (Uppercase) in the branch, and want to delete it...

Issue #2118, opened 2018-08-10
Clearing a bookmark note crashes the app


JBHemloque:
To repro:
- Create a bookmark and set a note
- Delete the note either by opening the bookmark dialog and deleting it, or deleting it inline in the bookmark grid

Note the app crashes.

I've got a fix for this, I'll submit a PR

robbyxp1:
Thanks!

PR #2119, opened 2018-08-10
Don't crash if the notes field is null

JBHemloque:
Fixes issues #2118 and #2120 

If a bookmark has a note and we clear it out, it will try to null it out. Set it to an empty string in this case, instead.

Also, if a bookmark's system cannot be found in the system DB, then we don't set the position in the bookmark editing dialog, which throws when we try to update the bookmark. The solution here is simply to fall back to the bookmark's position data if we can't find the system.


robbyxp1:
Thank you

PR #2139, opened 2018-08-28
Make CommandArgs debuggable; re-enable AppFolder in options.txt

klightspeed:
Getters should be idempotent.  The `Next`, `NextEmpty` and `Int` getters are not idempotent, and modify the state of the `CommandArgs` object.  Convert them to methods.

AppFolder processing from options.txt was also bypassed.  Re-enable it in a way that won't accidentally be bypassed again.


robbyxp1:
good point.. only was used in the netlogentry before this ;-)

robbyxp1:
Cheers

PR #2154, opened 2018-09-21
Add submodule update to appveyor.yml

klightspeed:
This will be required to get appveyor builds working with submodules


robbyxp1:
Thanks for this.

Recursive won't be needed I think.. I'm designing it so we don't recursively expand submodules of submodules, so you don't get two instances of baseutils for instance.

robbyxp1:
Thanks, done on branch

PR #2171, opened 2018-10-03
Prevent SLFs showing up in stats by ship

iainross:


iainross:
@robbyxp1 I made the change off the 10_2 patch branch so it will merge cleanly into that if you feel the need to do a release

iainross:
Added the hybrid fighters to IsFighter - still seeing a dozen jumps being picked up against a Type 6 and Type 7 and duplicated though.  There are a couple of duplicated IDs - a live Type 7, sold Type 6 and the Guardian fighter all have 18 in @Disorganise 's data which messes up counting stats by ship id.  I need to chase down where that comes from unless that's known (but not by me) which I won't get to until the weekend.

Probably want to hold off on merging this until then.

iainross:
@robbyxp1 this should be good now.

PR #2175, opened 2018-10-07
Record first discovery tags from SellExplorationData journals against the scans tagged

iainross:
The retrospective assignment of existing tags has been left as a menu option rather than forcing on start because it took nearly 15 minutes to trundle through all 70,000 of mine...


robbyxp1:
Holding due to volume of changes, will be brought in later

robbyxp1:
This one is out of date due to new EDDDiscoveryScan.. 

also it sort of violates the whole idea of the journal db having a pristine record of the journal entries, in json, and then we read and process them into the history list, accumulating information as we read it in. 

Why did you not do it like that?  

I think it needs a rethink to remove the changing of the db and just do it in memory..

iainross:
The pristine record thing is hardly inviolable, it's been done before (EDSM fields, map colour).

It's a shed load of string comparisons to tie the scans and data sales together, it's static information - once you have the tag you have it - and whether you got the tag is a logical extension of the scan itself.  All that made it feel to me like something worth doing once and persisting.

I could do it in memory but we'd be repeating an expensive operation over and over rather than leveraging the persistence we have with a database at the back of the application.

robbyxp1:
Yep, but we are paying the price for putting those extra fields in the DB when we send data to EDDN/EDSM etc.. having to remember to remove them.  And we (prob me) forgot to use a consistent naming strategy! 

So, first ther is a new journal record, MultiSellexplorationData which i think you need to process...

One thing i would consider doing is to split apart the computation of the tags from the normal flow of refresh history list..

In EDD form you could just call a function which processed the history to create the EDDTags.. reading the records..

The reason i say that is that the refresh history stuff is complex as it is.  I wonder if it would be logically cleaner to do that in its own space rather than this method.  it may not be.

And we need to make sure those EDD tags don't get to EDDN, so we need to check the tag removal is taking place.

I'll bring this in later in the flow, want to get a stable 10.3 out first.

iainross:
I'll need to look at the new journal, aye.  I'll have a look at moving it out of refresh history.

iainross:
Just sold some data in beta.  The new journal doesn't tell you what you tagged any more, just a number of bodies per system.  Fuck you, FDev.  They've holed this below the water line, I'll revisit if that changes for 3.3 release.

PR #2177, opened 2018-10-09
avoid crashing when no journal folder exists

jbvovau:


klightspeed:
`ExtraVersion.cs` is generated a build-time.  Please don't remove it from compilation.

jbvovau:
My bad, my pull was aimed to remove the crash when we want to display a journal directory that doesn't exist, but ExtraVersion.cs was previously removed

PR #2178, opened 2018-10-09
avoid crashing when no journal folder exists

jbvovau:
(ExtraVersion.cs keeped sorry for last error ;) )


robbyxp1:
Thanks - merged in in another way.

PR #2195, opened 2018-10-27
Habitable zone addon

spadino:
Added preliminary support for Water World, Earth like and Ammonia World orbital radius.


spadino:
Closing to push an updated PR

PR #2198, opened 2018-10-31
Circumstellar zones

spadino:
Added circumstellar zones properties for metal rich planets, water worlds, earth like worlds, ammonia worlds and icy planets.

Added context menu entries in Spanel for new zones, which can show/hide at will, and which states is saved in the db.


spadino:
Added global support for new **circumstellar zones**. Stars now report the new zones when highlighted in both Scan, ScanGrid and other panels which draw information from JournalScan.

Issue #2200, opened 2018-10-31
EDDN  heck MyReputation is being removed.


robbyxp1:
@here Please note that for now EDDN is silently removing the MyReputation part of  the Factions list as personal data,
following @OtisB (EDMC) PR, it will simply reject messages keeping it, date will be announced so software have time to update and user to migrate.
It should be around end of January 2019.

robbyxp1:
Done thanks Bravada.

PR #2203, opened 2018-11-07
prob #2200 Remove `MyFaction` from Factions in EDDN data

klightspeed:
Fixes 18441d8 - `MyReputation` is on the `Faction` entries, not the top-level event.


robbyxp1:
thanks.. i'll close the ticket

PR #2207, opened 2018-11-09
Only kick stardist on system change

klightspeed:
This should reduce the amount of freezing while reading events


spadino:
🤘🏻🤘🏻🤘🏻

Enviado do meu iPhone

Em 9 de nov de 2018, à(s) 06:30, Ben Peddell <notifications@github.com> escreveu:

PR #2211, opened 2018-11-11
Belts/Rings and ScanGrid improvement 

spadino:
- Added a contextmenu to ScanGrid, with various toggles
- Refactored the whole ScanGrid code, for a cleaner and improved parsing
- Added a new way to differentiate from Rings to Belts, and allow ScanGrid to parse them.


klightspeed:
Also, whatever tool is being used is making diffs noisy by converting spaces to tabs.

robbyxp1:
I think you need to pull this one and do it again.. please ensure your editor is set to space for tabs, not tabs.  Real tabs are bad.

robbyxp1:
Actually, more visible if its left opened.

spadino:
As I contribute to other projects which has different layout concept, sometime I forget to switch back the tab2tab feature... ;)

klightspeed:
Perhaps we should have an `.editorconfig` that sets the indentation type?

https://docs.microsoft.com/en-us/visualstudio/ide/create-portable-custom-editor-options?view=vs-2017

```
# top-most EditorConfig file
root = true

[*.cs]
indent_style = space
indent_size = 4
```

PR #2231, opened 2018-11-27
ScanGrid panel improved

spadino:
- added oneEarthRadius_m constant
- renamed solarRadius_m to oneSolRadius_m for terms consistency
- slight improvement in ScanGrid processing and display of bodies details (with belt and rings differentiation)

### DIFF
--- a/EDDiscovery/UserControls/UserControlScanGrid.cs
+++ b/EDDiscovery/UserControls/UserControlScanGrid.cs
@@ -136,119 +136,152 @@ namespace EDDiscovery.UserControls
 
             foreach (StarScan.ScanNode sn in all_nodes)
             {
+                // check for null data
                 if (sn.ScanData != null && sn.ScanData.BodyName != null)
                 {
+                    // define strings to be populated
                     StringBuilder bdClass = new StringBuilder();
                     StringBuilder bdDist = new StringBuilder();
                     StringBuilder bdDetails = new StringBuilder();
 
-                    if (sn.ScanData.PlanetClass != null)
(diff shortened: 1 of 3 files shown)


EDDiscovery/UserControls/UserControlScanGrid.cs
```diff
@@ -136,119 +136,152 @@ void DrawSystem(HistoryEntry he, bool force)
...
-                    if (sn.ScanData.HasRings && sn.ScanData.IsStar == false)
+                        // habitable zone for stars - do not display for black holes.
+                        if (sn.ScanData.HabitableZoneInner != null && sn.ScanData.HabitableZoneOuter != null && sn.ScanData.StarTypeID != EDStar.H)
+                            bdDetails.AppendFormat(Environment.NewLine + "Habitable Zone".Tx(this) + ": {0}-{1}AU ({2}). ", (sn.ScanData.HabitableZoneInner.Value / JournalScan.oneAU_LS).ToString("N2"), (sn.ScanData.HabitableZoneOuter.Value / JournalScan.oneAU_LS).ToString("N2"), sn.ScanData.GetHabZoneStringLs());                        
+                    }
+                    else
                     {
-                        if (sn.ScanData.Rings.Count() > 1)
+                        // is a non-stellar body
+                        if (sn.ScanData.IsStar == false)
```
klightspeed: This is superfluous (in this block we already know that `sn.ScanData.IsStar` is false), and wraps only the `if` statement below

spadino: Yes, indeed... ;)

PR #2237, opened 2018-12-07
Polish translation - additions and corrections

goscickiw:
Main file, user controls and journal events are complete, except for the things that I didn't know how to translate properly or in the right context. The ed file is left untranslated.

~~PS: Can somebody explain to me why are these changes to  ActionLanguage, BaseUtilities and ExtendedControls in my commit? I didn't do anything with them.~~ I have recreated the commit without adding these files, they caused the checks to fail.


spadino:
Well done, @goscickiw! Thanks!!

PR #2239, opened 2018-12-10
Feature/scangrid

spadino:
* Fixed missed distance for planets and moons (last update bug)
* Added a context menu with toggles for some information field and column, which state is stored and retrieved from DB.


robbyxp1:
You may want to update the help on the wiki on this.. its not obv to click the top menu bar to get the context menu..

PR #2245, opened 2018-12-12
Feature/scangrid

spadino:
- Added a statusbar at the bottom
- Moved the estimated total value to the status bar
- Included a summary in the control title, after the system name, which tell how many known stars, planets (terrestrial and gas giant) and moons are in the system
- Add an animated progress bar which graphically shows how many jumponium materials are available in the system
- If a system has ALL jumponium materials, a GREEN SYSTEM label (with a green background) shows up in the status bar


robbyxp1:
Looks good.. 
Its not themed correctly, but we can fix that up later..

robbyxp1:
The green back png is not required.. can't you use the BackColor property of the statusGS label to colour it..

that would reduce file changes..

PR #2254, opened 2018-12-14
Process ring scans

klightspeed:


robbyxp1:
I've coded it already - will cross check

PR #2255, opened 2018-12-14
Feature/scangrid

spadino:
Some fixes:

- Fixed statusbar; not need for green_back.png
- Belt Clusters now has custom icon
- Blet Cluster can be hide or show, via context menu


robbyxp1:
This will have to wait.. too much change for this release.

robbyxp1:
Actually, its not as large as it looks. sold.

Issue #2257, opened 2018-12-15
Wrong total (amount + bonus) on Sell Exploration Data


EnterTheNameHere:
I've sold eploration data at Universal Cartographics, but there seems to be an issue with the total amount. Here is the information from EDDiscovery:
https://imgur.com/tLHWoOr

Here is entry from Codex:
https://imgur.com/PisLc15

And here is Copy journal entry to clipboard/Log
```
{
  "timestamp": "2018-12-15T07:26:49Z",
  "event": "SellExplorationData",
  "Systems": [
    "58 Geminorum"
  ],
  "Discovered": [
    "58 Geminorum BC 1",
    "58 Geminorum BC 9",
    "58 Geminorum A 2",
    "58 Geminorum BC 6",
    "58 Geminorum A 9",
    "58 Geminorum A 5",
    "58 Geminorum BC 7",
    "58 Geminorum BC 5",
    "58 Geminorum A 8",
    "58 Geminorum BC 8",
    "58 Geminorum BC 4",
    "58 Geminorum A 10 A Ring",
    "58 Geminorum A 3",
    "58 Geminorum BC 3",
    "58 Geminorum A 4",
    "58 Geminorum A 7",
    "58 Geminorum A 1",
    "58 Geminorum A 6"
  ],
  "BaseValue": 9858271,
  "Bonus": 9450644,
  "TotalEarnings": 10058271
}
```

EnterTheNameHere:
As you can see, the total amount EDDiscovery thinks is sum of BaseValue + Bonus, but Total Earnings are not the sum. The wrong amount is then reported to INARA etc.

klightspeed:
EDDiscovery uses `BaseValue + Bonus` when `TotalEarnings` is lower due to a bug in E:D 3.0.2 (where TotalEarnings was much lower).

klightspeed:
Just how many earthlikes or water worlds did that system have?  And the `Discovered` says you should have earned a bonus.  Yet the difference between `BaseValue` and `TotalEarnings` is 0, suggesting either `BaseValue` is bogus or you got no first discovery bonus.

Looks like we need to alter the test to only try to correct the error if the event was from before e.g. April 2018.

EnterTheNameHere:
58 Geminorum is:
2 high metal content terraforming candidate I discovered and mapped first (according to in game system map)
10 high metal content (non candidate) I discovered and mapped first
1 high metal content being terraformed (not me)
[...]

PR #2258, opened 2018-12-15
Prob 2258 Only correct TotalEarnings if before May 2018

klightspeed:
This should work around the journal error in #2257


robbyxp1:
Ta

PR #2259, opened 2018-12-15
Allow the a waypoint have the same distance as a previous one. 

### LINKED ISSUE #2252: Route Tracker does not advance to next Point
I plotted a Route with the Neutron Router and imported it into EDDiscovery via Text File.

Those Systems are unknown to EDSM (I downloaded only Bubble + Colonia). 
Now in the first unknown System, it wants to set destination to the previous system.

Route:
- Prua Dryoae ZQ-M a34-3
- Pro Eurl GL-P d5-20
- Prua Dryoae AB-U d4-39
- Prua Dryoae IC-S d5-1
- Prua Dryoae LN-Q d6-18
- Pro Eurl VX-L d7-19
- Swoiphs CT-O d7-72 (unknown)

During the last 3 jumps ("Prua Dryoae IC-S d5-1" to "Pro Eurl VX-L d7-19") it did not advance to the next waypoint and said 0.0 Ly to the next way-point.

As a workaround for both Problems I would suggest some other behavior:
When arriving in a system: if the current system is found on the route set the way-point to the entry directly after it.

To avoid problems with multiple entries of the same system: look for the system name on the route on entries behind the last visited way-point.

daniel-finger:
After some testing this works very well for me.
Fixes #2252

### DIFF
--- a/EliteDangerous/DB/SavedRouteClass.cs
+++ b/EliteDangerous/DB/SavedRouteClass.cs
@@ -397,7 +397,7 @@ namespace EliteDangerousCore.DB
                     //System.Diagnostics.Debug.WriteLine("From " + knownsystems[i - 1].ToString() + " to " + knownsystems[i].ToString() + " Percent " + interceptpercent + " Distance " + dist);
 
                     // allow a little margin in the intercept point for randomness, must be min dist, and not stupidly far.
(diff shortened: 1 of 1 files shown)


robbyxp1:
Welcome aboard.

PR #2266, opened 2018-12-18
Fix filename case in designer files

klightspeed:
Incorrect filename case is causing the Travis built to fail.


robbyxp1:
One thing i dislike about unix..

PR #2275, opened 2018-12-20
 Fix log fetcher and system fetcher rate limits

klightspeed:


robbyxp1:
Just checking with Anthor about why his server would have changed..

PR #2282, opened 2018-12-21
Fixed Total Fired count in crew statistics

goscickiw:
Fixed Total Fired count displaying the same as Total Hired count. Compiled and tested by hiring and firing another NPC. Seems to work fine.


robbyxp1:
thanks

PR #2288, opened 2018-12-28
Make UCs populate when travel history is completely filtered

iainross:
Makes most recent HE the default if everything is filtered out, rather than no change at all being notified to dependent UCs (or no initial load happening if loading fully filtered by a time window)


robbyxp1:
Thank you.

PR #2296, opened 2019-01-02
Update Search for searching scan data..

robbyxp1:
Add search scan tab.. update action doc.. move base utils/ext controls/action lang on


spadino:
Fantastic!

Issue #2304, opened 2019-01-05
Statistics Travel tab 10.3.3 not calculating 'all' column correctly


Disorganise:
Per the screenshot, some of the items in the 'all' column (Jet cone boosts, heat warning/damage, fuel scooped, scooped tons) are equating to the last docked.   Even those numbers may not be correct - they look post 3.3 only maybe.
![image](https://i.imgur.com/I8RUC8G.png)

iainross:
Mine all look right - have you got an old entry highlighted in the history tab or has it got a date filter set?

What have you got selected on the settings tab for Only Read Essential Entries?

Disorganise:
Nope to first question - I restarted the app also so if I *did* have something inadvertently set, it should (hopefully) get forgotten.
to 2nd:  >7 days.  Is that it?  I've been gone for about 2 weeks :/

iainross:
That would appear to be it, if I set mine to >7 days then the same entries are based on incomplete data.

robbyxp1:
Yep, nothing much we can do about that.  Its a choice between memory / full data at the moment.

Disorganise:
ok, I'll put it back to long, at least when I want to read the stats.  I'm sure there was a performance reason I restricted it (not memory, got heaps of that - cpu cycles is my limit)

robbyxp1:
Iain, as some point, we may want to get the stats for certain parts from the DB instead of the history list, but thats a long term thingy.

iainross:
I was thinking about maybe a 'Full stats' option for the Essential journals drop down in 10.4 that's the current essential plus anything extra the Stats panel looks at.

Longer term collating them in a database table is probably a good idea though.

robbyxp1:
Yep, that would be a good idea. We can easily put more options in there.

The DB option is a bit of a jump , so thats prob a better idea for now.

iainross:
I'm in the middle of adding a mapped bodies count at the minute, so I'll do it at the same time.

robbyxp1:
Iain has added a full stats option to 10.4, i think this is closed.

PR #2305, opened 2019-01-05
Count Mapped Bodies in statistics panel, add option to include all stats journals in Essential entries

iainross:
Solves #2304


robbyxp1:
In.

PR #2310, opened 2019-01-09
Add meaningfull overlays to scangrid icons.

spadino:
Add some overlays to scangrid icon column:
- landable
- vulcanism
- materials
- surface mapped

image

---

> **In lack of a better solution, I used the ErrorText to pass some needed information. Could someone point me to a better solution?**


klightspeed:
Since it's in `cur.Cells[0]`, why not have a structure in `cur.Cells[0].Tag`?

robbyxp1:
Yep, just have a struct { bool volcanism; ..} and use that.

I'll hold and let you consider.

PR #2315, opened 2019-01-10
Only send rankValue if it has a value

### LINKED ISSUE #2314: Sending invalid parameters for Engineer ranks to Inara.
EDD is sending a rankValue parameter to Inara for Engineers that are not unlocked.

This marks the engineer as unlocked on Inara's website, despite them not being unlocked.

rankValue should not be sent for engineers that are not unlocked.

For example, it's sending this for me:

{
      "eventName": "setCommanderRankEngineer",
      "eventTimestamp": "2019-01-10T20:04:10Z",
      "eventData": {
        "engineerName": "Etienne Dorn",
        "rankStage": "Invited",
        "rankValue": 1
      }

And that marks the Engineer as unlocked, due to the "rankValue": 1.

According to their API documentation (https://inara.cz/inara-api-docs/#event-9): Note: 'rankStage' and 'rankValue' properties aren't required, but there should be at least one property set. When the 'rankValue' is set, engineer will be in a stage 'Unlocked', doesn't matter what was set.

klightspeed:
This should fix #2314

### DIFF
--- a/EliteDangerous/Inara/Inara.cs
+++ b/EliteDangerous/Inara/Inara.cs
@@ -170,13 +170,14 @@ namespace EliteDangerousCore.Inara
             return Event("setCommanderGameStatistics", dt, jsonfromstats);
         }
 
-        static public JToken setCommanderRankEngineer(string name, string progress, int value, DateTime dt)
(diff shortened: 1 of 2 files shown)


robbyxp1:
Already had this done.. thanks.

PR #2329, opened 2019-01-16
3.3 scan values

iainross:


robbyxp1:
Thank you. Will pull in tomorrow when i'm back on it.

PR #2330, opened 2019-01-16
Fix some wrongly assigned overlays

spadino:
I found that it's sometime better to re-scan the journals...


robbyxp1:
tomorrow ;-) getting tired!

PR #2336, opened 2019-01-17
Fix EDSM button tooltip; confirm sync when sync disabled

### LINKED ISSUE #2334: EDSM Sync Button on Toolbar ignores configured Settings
If I press the `EDSM Sync` Button on the Toolbar it uploads data to EDSM, even if I specifically configured EDDiscovery to only fetch data from EDSM.

Even if the Tooltip Text suggests that it does a TO and FROM Synchronization, I still expect it to respect my configured settings.

klightspeed:
This should clarify the purpose of the EDSM button, and fix #2334


robbyxp1:
Going to take the code, but you put in synctoEDDN due to a typo. - so i'll just copy it in and reject the push!

Rob

PR #2361, opened 2019-01-20
Changed default rows height for improved bodies icons

spadino:


robbyxp1:
The planet landing icon need tlc as well..

PR #2371, opened 2019-01-21
ScanGrid features: landable overlay icon; orbital details

spadino:


robbyxp1:
Master is not open for enhancement, only for bug fixes. This will have to wait until the next cycle starts, which won't be for at least two weeks.

robbyxp1:
Push is incorrect.

EDDIcons.csproj why is the pngs being removed? Why is touchdown not in there..

scan grid - make doubly sure that any properties you use have been checked for null - be paranoid.  That caused a crash in 10.4. I'm sure you have but double check

So, in the icon drawer, you replaced int with var. Why?  You should always choose the most typed representation of the object you can.  var should be reserved for situations where you have a crazy definition (Tuple comes to mind)..  int's dont really classify as crazy.  It hides the underlying type (int vs long?).  

Its also masking the actual changes you made.

have another go! ;-)

PR #2376, opened 2019-01-22
Prob 2370 Call Closing if not already called while disposing

### LINKED ISSUE #2370: Error: Unhandled UI Exception - NullReferenceException
### Description
I was doing a first-time run of the application and this error did it's error thing.

klightspeed:
This should fix #2370


robbyxp1:
I'm going to implement this, but in a slightly simpler manner, mainly because i'm worried about forgetting to call base.Closing() in any new code.  It also means only 7 files need chaning.

Basically get the callers to CloseDown() which calls Closing.  Nobody except CloseDown calls Closing() that means that the base. call does not need to go into the code, and means only a few files need changing.

PR #2394, opened 2019-01-26
Created a small overlay which show worth to be surface mapped bodies.

spadino:
Guys, here the work I've done so far for a very small overlay, to shows bodies which are worth to be surface mapped (user definable).

By now, I can generate a list of bodies which satisfy the properties defined in the context menu. 

I need some help to deal with the GUI, as I'm not able to draw to the picturebox, no matter what I do. I've already tried many solutions from other overlays, but none seems to work... also, the overlay is not able to be transparent, which of course is a problem ;)


spadino:
Done! ;)

PR #2400, opened 2019-01-28
Created a small overlay which show worth to be surface mapped bodies.

spadino:
It display a list of bodies which fulfill some very strict requirements; the purpose of this panel is to aid during exploration, so it notify which bodies are worth to be mapped.

---

Planned features:
- add more bodies classes, for more generic searches
- add notable stellar phenomenon


spadino:
Thanks! Actually, GUI interaction was just a placeholder, it wasn't implemented yet... ;)

Issue #2402, opened 2019-01-29
Incorrect jump range calculation in “System Information” panel


kaimi:
So both my Beluga and my ExplorEagle show up as way less range than they actually have; the difference seems to be pretty close to their fitted FSD boosters.

![image](https://user-images.githubusercontent.com/910302/51914148-7424fd00-23d8-11e9-9b02-927ee6bbf14d.png)

![image](https://user-images.githubusercontent.com/910302/51914189-8bfc8100-23d8-11e9-8a34-8a7016249af9.png)

bdeferme:
Same issue here, very close to the FSD booster range.

robbyxp1:
FSD Boosters are not in the calculation at the moment.

robbyxp1:
In code base, thank you

kaimi:
Thanks for the quick fix :)

PR #2410, opened 2019-02-02
Use FSD Booster in Jump Range Calculation

robosa:
Fix for Incorrect jump range calculation in “System Information” panel #2402


robbyxp1:
Thanks. I'll pull this in later this week.

next ED is scheduled for later in the month.

PR #2419, opened 2019-02-07
Fix build error in tests due to interface change

klightspeed:


robbyxp1:
Darn, always forget the tests, thanks

PR #2434, opened 2019-02-21
Add bodies count after discovery scan.

spadino:


robbyxp1:
sn could be null, you need to protect ;-)

klightspeed:
Why reference the system node when you have the FSSDiscoveryScan from which TotalBodies is set?

spadino:
@robbyxp1 , I included a null check.

@klightspeed, what you suggest? How would you do that?

klightspeed:
>> Why reference the system node when you have the FSSDiscoveryScan from which TotalBodies is set?
>
> what you suggest? How would you do that?

```
var je = he.journalEntry as JournalFSSDiscoveryScan;
if (je != null)
{
    var bodies_found = je.BodyCount;
    SetControlText(bodies_found + " bodies found.");
}
```

EDDiscovery/UserControls/Overlays/UserControlSurveyor.cs
```diff
@@ -126,9 +120,19 @@ private void NewEntry(HistoryEntry he, HistoryList hl)
...
             {
+                SetControlText(string.Empty);
                 pictureBoxSurveyor.ClearImageList();
                 Refresh();
-            }            
+            }
+
+            if (he.EntryType == JournalTypeEnum.FSSDiscoveryScan)
+            {
+                if (sn.TotalBodies != null)
```
klightspeed:
```
sn?.TotalBodies
```

PR #2441, opened 2019-02-27
Add user definable text alignment (left, center, right)

spadino:
It allow Surveyor to list bodies in left, center or right text alignment.


robbyxp1:
All of this is already built into the baseutilities : DrawTextInfo functions. Your repeating the code.. have a look at that - it can use StringFormat to indicate alignment inside a bitmap box.

robbyxp1:
And use that StringFormat class to hold the alignment.  Those functions are proven to be working.

PR #2442, opened 2019-02-27
Feature/alignment

spadino:
Allow Surveyor to align body list to left, center or right.


robbyxp1:
So, the thing is, is that your using a bitmap, and drawing into it to find its bounds...

but that's all done by  BaseUtils.BitMapHelpers.DrawTextIntoFixedSizeBitmapC with using either center = true, or centre = false and giving a string format.

I'll take it in but may optimise it later ;-)

spadino:
Yeah, we can optimise it later, for sure. I need to include some other notable bodies features (has guardian ruins, etc...) so I will try to improve on that by myself. But, of course, you are the guy, it must be acceptable for you. Thanks for your advises and kindness, as always. It’s a real pleasure and a great opportunity to deal with you. I’m learning so much...
Cheers.

Enviado do meu iPhone

Em 28 de fev de 2019, à(s) 04:49, Rob <notifications@github.com> escreveu:

PR #2445, opened 2019-03-01
Chinese translation

Nats-ji:


spadino:
Many thanks, @nats-ji

PR #2473, opened 2019-04-10
SysInfo transparency restored to pre-scrollbar level of functionality

iainross:
I like to keep System and Visits floating in the top right and the ExtScrollPanel stuck a big black box round them.  It's far from perfect as an overlay but for a couple of fields it works OK.


robbyxp1:
966 looks a bit suspicious.. ;-)

i'll check, thanks

robbyxp1:
fixing outside of pull, thanks

Issue #2457, opened 2019-03-20
 Cannot add scan to system


Andromedadragon:
I'm getting the message, cannot add scan to system with the current system name and object that was just scanned by FSS. I included the log file that Elite created
[Journal.190320175311.01.log](https://github.com/EDDiscovery/EDDiscovery/files/2990298/Journal.190320175311.01.log)

robbyxp1:
Thanks - will try that and see if i can reproduce

klightspeed:
What version of EDDiscovery are you using?  Does the trace log (in `%LOCALAPPDATA%\EDDiscovery\Log` or `Data\Log` for the portable version) indicate which body it is unable to process, or is it occurring with all bodies?  Does the trace log indicate any errors processing the `FSDJump`?

Andromedadragon:
I'm using 10.5.2.0

It was all bodies in the one system.

I included the log files that were in the EDDiscovery file. Please note that I did try to recreate the database thinking that it might have corrupted since I hadn't used in a few days  or the system didn't shutdown correctly. I dropped in the respective log files from around that time period. I wish that I included the time and date in my original, so that I could include only the specific ones.
[Trace_2019-03-20-17-58-25.0.log](https://github.com/EDDiscovery/EDDiscovery/files/3011077/Trace_2019-03-20-17-58-25.0.log)
[Trace_2019-03-20-17-59-27.0.log](https://github.com/EDDiscovery/EDDiscovery/files/3011078/Trace_2019-03-20-17-59-27.0.log)
[Trace_2019-03-20-18-01-07.0.log](https://github.com/EDDiscovery/EDDiscovery/files/3011079/Trace_2019-03-20-18-01-07.0.log)
[Trace_2019-03-20-18-05-42.0.log](https://github.com/EDDiscovery/EDDiscovery/files/3011080/Trace_2019-03-20-18-05-42.0.log)
[Trace_2019-03-20-18-16-37.0.log](https://github.com/EDDiscovery/EDDiscovery/files/3011081/Trace_2019-03-20-18-16-37.0.log)
[Trace_2019-03-20-17-54-19.0.log](https://github.com/EDDiscovery/EDDiscovery/files/3011082/Trace_2019-03-20-17-54-19.0.log)

Andromedadragon:
I had it occur again. I've added the files below.
[...]

PR #2475, opened 2019-04-12
Throw away the buffer on I/O error or reading nulls

klightspeed:
This should go some way to fixing #2457 (though we still need to tell the OS to invalidate its cached copy of the file).


robbyxp1:
Thanks

PR #2480, opened 2019-04-23
Eliminate use of DataTable in JournalEntry.GetAll

klightspeed:
This should reduce the startup memory usage with large histories.


robbyxp1:
Wow, that one has been left in a long time.

In the new db, those SQLQueryText's are going to be eliminated i think.

thanks

klightspeed:
Yes - PR #2481

PR #2500, opened 2019-05-19
Correct 3.3 ELW value calculation

### LINKED ISSUE #2499: Incorrect Earth-like estimated value
The calculation of the estimated value of earth-like seems a bit off.

The majority of the values and formulas are found [here](https://forums.frontier.co.uk/threads/exploration-value-formulae.232000/)
The formula for the calculation is k + k * q * mass^0.2 and this is used correctly for every other planet type.
The calculation for earth-likes seems to have the wrong k value.

To elaborate this a bit more:
I'm currently at Rohini. This system has one earth-like: Rohini AB 1
The exported estimated value of EDDiscovery is 171,944
With the mapping bonus etc. EDDiscovery estimates a value of 716,436 (Seen [here](https://imgur.com/u1GxmCx))
Most likely calculated like this:
`171,944 * 3.3333333333 (mapping multiplier) * 1.25 (efficiency bonus) = 716,433.3333`

This would be correct for a k value of 116295
`116295 + 116295 * 0.56591828 * 0.4323^0.2 = 171,945.8514`
 
But that's the only the terraformable part of k.
For earth-likes it is the same as a terraformable water world.  Which is `64831 + 116295 = 181,126`

To back this up with actual in game data:
[Here](https://imgur.com/wdiJ6X7) is a screenshot of the exploration values for the Rohini system
It lists the earth-like with a value of 1,115,825
This is with the mapping and efficiency bonus. Calculated like this:
`181,126 + 181,126 * 0.56591828 * 0.4323^0.2 = 267,800.5441`
`267,800.5441 * 3.3333333333 (mapping multiplier) * 1.25 (efficiency bonus) = 1,115,835.6`
That is only 10 off. 
As seen above currently estimated value is:
716,436
That is roughly 400,000 off

iainross:
Fixes #2499

### DIFF
--- a/EliteDangerous/JournalEvents/JournalScan.cs
+++ b/EliteDangerous/JournalEvents/JournalScan.cs
@@ -1216,7 +1216,8 @@ namespace EliteDangerousCore.JournalEvents
                             if (Terraformable) kValue += 116295;
                             break;
                         case EDPlanet.Earthlike_body:
-                            kValue = 116295;
+                            // Always terraformable so WW + bonus
+                            kValue = 64831 + 116295;
                             break;
                         default:
                             kValue = 300;


robbyxp1:
Ta.

PR #2504, opened 2019-05-24
Add stellar bodies count

spadino:


robbyxp1:
I'm in the middle of a big rewrite of these panels to handle high DPI.. i'll bring this in in a while..

thank you

robbyxp1:
this is out of date,, it needs a rework from the work i did to fix and stabilise the surveyor after a few bug reports.. i suggest you rework and resubmit. thanks

spadino:
Ok! Thanks

PR #2515, opened 2019-06-05
Fill scan mappings in CSV export

### LINKED ISSUE #2514: CSV export not showing mapped values
The CSV export triggered from a SCAN tab does not include full values for mapped planets. it only shows the estimated value for the planet as if it was not mapped. 
included is an image showing the difference between the CSV reported values, and EDDiscoveries values: https://cdn.discordapp.com/attachments/357165142430973952/585830910809735168/unknown.png

klightspeed:
This should fix #2514

### DIFF
--- a/EDDiscovery/UserControls/ScansStars/UserControlScan.cs
+++ b/EDDiscovery/UserControls/ScansStars/UserControlScan.cs
@@ -634,6 +634,16 @@ namespace EDDiscovery.UserControls
                             {
                                 JournalScan scan = je as JournalScan;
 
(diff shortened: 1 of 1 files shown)


robbyxp1:
Okay, i think there is an IsStar function btw if thats what your getting at checking the planet class..

klightspeed:
Do rings have an estimated value?

robbyxp1:
So, to read past your comment, your using planet class to determine its not a belt or something else.

It would be better to put an explicit IsPlanet function in the scan class to indicate its not a star or a belt or anything else, and then the code above would be clearer that its a planet you want.

In the meantime, its in.

PR #2517, opened 2019-06-06
Wrap SQL database access; move to its own thread

klightspeed:
This should finally fix the database locked errors by ensuring the database is only accessed from a single thread.

Also debug any instances where the database is opened twice (e.g. accessing registers while holding an open connection).

I will rebase onto master once ready.


robbyxp1:
Thats a very big change. 

Its not going to be in 11.0 with the amount of time left.  I've been testing this for a month now in its current iteration and its stable.  I only have about 2 weeks of effort left before 11.0 and changing something of this magnitude is not going to be possible.

robbyxp1:
I suggest you hold this in your back pocket for post 11.0

robbyxp1:
Can you update this to the master, I think it needs resyncing and it prob would be best if you did it.

thanks

PR #2520, opened 2019-06-12
Override temp directory

klightspeed:
This should prevent the SQLite library from writing to the user's temp directory.


robbyxp1:
I'm torn on this one.  Because a user may have explicitly set up TMP/TEMP folder to another place to minimise usage on their primary SSD.  By doing this, we are overriding their selection.

We need to think about this.. is this the right thing to do?  Its only 1 user who reporting any issues on this.

klightspeed:
Perhaps have an option to override the temp directory?

PR #2528, opened 2019-07-05
Fix empty check from 2ca6bf4

klightspeed:
Fixes: 2ca6bf4b7c816a300aab674d5c1a56e96deb783e Ensure Systems and Discovered are not null


robbyxp1:
Later, One other thing we could do would be not print those empty "" fields..

PR #2549, opened 2019-08-02
UI shopping list

spadino:
Add a toggle which stack the shopping list horizontally, for better screen estate usage on vertical monitors.

image

image


spadino:
Closing. Pushing a cleaner PR in a while...

PR #2550, opened 2019-08-02
Converted the Settings panel in a tabbed flow controlled scheme. Settings groups are now more compact, and they will be repositioned to fit smaller screens.

spadino:
Split settings in Settings panel in a tabbed interface (basic and advanced setup) and flow controlled position of them.


spadino:
Closing. Improve  idea the after testdrive

PR #2552, opened 2019-08-05
UI settings panel design further improvements

spadino:
Improve on the new PanelScroll flow behavior, providing a tidy settings panel, which allow for clean shows on client size from the theoretical minimum of 768 (for the smaller vertical monitor currently available) up to wide and extra wide monitors.


robbyxp1:
So,

I like it,

try it with some other languages.. i think its too compressed to work, for instance under the Memory section - there is not enough space for the other languages between the text and the button.

Have a try and space it out and then we will pull it in.

rob

robbyxp1:
I would make the memory section larger, with the text on one line, button underneath.  Also consider expansion of text for all the other items.

I've learnt to leave a lot of space for text expansion for english vs other languages.

spadino:
Sure! Thanks for your feedback, Robby.

Obter o Outlook para Android<https://aka.ms/ghei36>

________________________________

robbyxp1:
Please withdraw this if you don't want it in here, don't want it hanging around as it will inevitably get out of date.

spadino:
image

PR #2554, opened 2019-08-08
Unsubscribe from subscribed events when closing

klightspeed:
This should fix some crashes due to events being handled after the controls are disposed.


robbyxp1:
Thank you!

PR #2578, opened 2019-09-09
Fix race when adding system to cache

klightspeed:


robbyxp1:
Thanks, missed that one

PR #2581, opened 2019-09-11
Prevent divide by zero in PanelSelector control

klightspeed:


robbyxp1:
That old one.. thanks

PR #2583, opened 2019-09-14
Don't remove the default directory from the watch list

klightspeed:
This should fix the issues where EDDiscovery stops seeing new events when commanders override their journal directory.


robbyxp1:
Thanks

PR #2598, opened 2019-09-24
Update scan display when materials are collected etc.

### LINKED ISSUE #2597: Material Collection not updating on scan tab
Hello, I recently upgraded from 10.6.0 to 11.0.3.0 and I've just starting to collect materials after about 40lys of travel. I can see from the journal view that the materials are being registered in EDD, but the scan window with 'show only required materials' selected, does not reflect the additional material. The white popout window stays the same value, and if I max out the material, the hex icon is supposed to be removed but it doesn't. Now I've collected several times and it never updated. I've restarted and it's corrected the materials on loading, but still fails to update on further collections. I tried again today and it's still the same. I'm looking at a re-install next. This used to work fine on the previous version. Is there anything I can do? Thanks in advance.

klightspeed:
This should fix #2597

### DIFF
--- a/EDDiscovery/UserControls/ScansStars/UserControlScan.cs
+++ b/EDDiscovery/UserControls/ScansStars/UserControlScan.cs
@@ -145,7 +145,7 @@ namespace EDDiscovery.UserControls
         {
             if (he != null)
             {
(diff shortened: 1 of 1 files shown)


robbyxp1:
cheers, may just copy it into a push tomorrow which does even more to scan

robbyxp1:
Manually pulled into my push, thanks

PR #2605, opened 2019-09-29
Use friendly body type names in Stats panel

klightspeed:


robbyxp1:
Okay, but..

Why are we repeating some of the types?  As you said, Starname is in there..  its translatable.  Would we not want to show those?

robbyxp1:
The reason i ask of course is because it should be translated, if possible, and its a duplication of information.. sort of

robbyxp1:
cheers

PR #2612, opened 2019-10-02
Add SAASignalsFound to events sent to EDDN

klightspeed:
EDSM-NET/EDDN@3eed3bc added SAASignalsFound to the events accepted by EDDN.


robbyxp1:
thanks

PR #2615, opened 2019-10-04
Eliminate unnecessary system DB connection

klightspeed:
The unnecessary system DB connection results in a double-open


robbyxp1:
Okay, Thanks. i'm preparing a user DB push over the weekend, and will bring this fix in as well.

PR #2618, opened 2019-10-05
Add FSD boosts to the route finder

EoD:
This PR adds FSD Injection boosts to the route finder in case it is unable to find a system.

Here is an example:

* Without FSD boosts edd_before

* With FSD boosts edd_after

In the latter image, I have the target system twice in there. This is kind of intentional as I wasn't really sure how to visualize the need for FSD boosts on the target.


klightspeed:
Of course the reason there mightn't be known systems to jump to might be because nobody's been there, and jump boosts might not actually be needed.

EoD:
@klightspeed I agree that it doesn't make sense everywhere, but in case of often traveled routes (like the Abyss) it is very helpful imho. On the example route in particular, most of the stars route are already discovered. Unfortunately the stars are just too far apart for the game to find a route successfully and the above PR was my solution for this scenario.

klightspeed:
Perhaps include jump boosting as an option in the metric dropdown or only if the requested range is less than e.g. 100ly - if someone is using a 1000ly range in order to e.g. plot economical routes, they probably don't want some of those waypoints to be 1500ly or 2000ly apart if it's through an unexplored area of the galaxy.

EoD:
@klightspeed I added a checkbox (suggested by Robby) which allows toggling FSD boosts on or off. Does this sort your concerns?

EoD:
Marking this as "ready for review" as my personal blockers have been resolved.

robbyxp1:
Thanks!

PR #2626, opened 2019-10-11
Fix inappropriate uses of ToStringInvariant

### LINKED ISSUE #2625: distances in search tab are not sorted correctly
http://prntscr.com/pi0krr

klightspeed:
For strings that are displayed directly in the UI, or are sorted using double.Parse, using double.ToStringInvariant is inappropriate, and for sortable columns results in broken sorting.

This should fix #2625


robbyxp1:
thanks

PR #2631, opened 2019-10-14
Prevent showMaxInjections being overridden by saved setting for toggl…

iainross:
…eShoppingListPosition on load of UserControlShoppingList


robbyxp1:
good spot.

PR #2635, opened 2019-10-19
Prevent deadlock when closing while route is being computed

klightspeed:


robbyxp1:
We will have to solve it in another way. BeginInvoke is a bad move from the UI point of view. Thats why I changed it.

klightspeed:
Refer https://referencesource.microsoft.com/#system.windows.forms/winforms/managed/system/winforms/control.cs,7841

`Control.Invoke` calls `WaitForWaitHandle(tme.AsyncWaitHandle)`, while `Control.BeginInvoke` returns `tme` where we ourselves can call `tme.AsyncWaitHandle.Wait()` (or wait for either that or the cancellation event).

PR #2643, opened 2019-10-22
Avoid ambiguity by using an enum for route metric

EoD:
With the previous approach it was very hard to make sure that a selection from the ComboBox is guaranteed to be the selected metric.

This PR is kind of a helper for a follow-up PR of mine.

I moved the enum definition into `RoutePlotter` as this was the best place I could think of. Any suggestions?

Before this PR the enum was once in `EliteDangerousCore.RouterPlotter` and once in `EliteDangerousCore.DB.SystemsDB` as it was needed by `SystemsDB.GetSystemNearestTo()`. From my naive point of view, these constants have nothing to do with a DB though.
`GetSystemNearestTo()` could also easily be moved into any other place as it has no dependency on the DB or anywhere else, except the new enum. The best way might be to move this method into maybe a helper class which collects all "systems calculations/metrics" and some plotting helpers. Is it worth the refactoring?

### DIFF
--- a/EliteDangerous/EliteDangerous/RoutePlotter.cs
+++ b/EliteDangerous/EliteDangerous/RoutePlotter.cs
@@ -17,6 +17,7 @@
 using EMK.LightGeometry;
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 
 namespace EliteDangerousCore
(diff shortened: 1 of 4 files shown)


robbyxp1:
Looks good, except the switch statement for the text can be replaced by a const string[] array.  A switch statement looks a bit odd.  You can use an assert beforehand to check the enum range.

EoD:
@robbyxp1 I don't quite understand what you mean. Can you sketch the code?

At the beginning I had an if construct, but the switch looked much cleaner. The reason I used this construct is that I need to assign a string to a generically-typed enum value and it is (afaik) impossible to make sure that those match with their respective counterpart if I were to use something like an array.
You can never guarantee which enum has which index in case of an extension of the array. That's why I am using an `foreach` in the `UserControlRoute.cs` instead of the original `for`.

I accepted the drawback, that what ever is saved (and hence loaded) in the database might not be 100% consistent with the current code. The consequences are minor at best. Even now, it is not guaranteed that the saved value is identical to the existing list in the code.

EliteDangerous/EliteDangerous/RoutePlotter.cs
```diff
@@ -29,18 +29,40 @@ public class RoutePlotter
...
         // METRICs defined by systemclass GetSystemNearestTo function
-        public static string[] metric_options = {
-            "Nearest to Waypoint",
-            "Minimum Deviation from Path",
-            "Nearest to Waypoint with dev<=100ly",
-            "Nearest to Waypoint with dev<=250ly",
-            "Nearest to Waypoint with dev<=500ly",
-            "Nearest to Waypoint + Deviation / 2"
-        };
+        public static string GetMetricName(Metric metric)
```
robbyxp1: This needs replacing with a string[] and using the enum to look up the text.  Since the class has the enum, its safe to just use a string[] to find the text to go along with the enum

EliteDangerous/EliteDangerous/RoutePlotter.cs
```diff
@@ -29,18 +29,40 @@ public class RoutePlotter
...
+                case Metric.IterativeMaximumDev500Ly:
+                    return "Nearest to Waypoint with dev<=500ly";
+                case Metric.IterativeWaypointDevHalf:
+                    return "Nearest to Waypoint + Deviation / 2";
+                default:
+                    throw new ArgumentOutOfRangeException(nameof(metric));
+            }
+        }
+
+        public enum Metric:int
```
robbyxp1: Metric:int you don't need the int. Its that by default.

EoD: Removed the `int`

robbyxp1:
I would be tempted to move the enum for metric into SystemDBGetByDistance.cs file, and make it own it.  Then its associated with the code which needs the metric, and then you can use it further up the chain.

EoD:
From my naive point of view, these metric constants have nothing to do with a DB though and I would instead move this metric function out of the DB code there.

Especially in hindsight that there is much more metric-specific code going to come which has nothing to do with the DB as well:
https://github.com/EoD/EDDiscovery/blob/add_astar_routing_method/EliteDangerous/EliteDangerous/RoutePlotter.cs#L441-L469

Why do you want the metric to be in the DB section of the code?

EoD:
@robbyxp1 where do you want it to have? In the Plotter or in the DB?

robbyxp1:
The blocker is i'm busy with other stuff ;-) Be doing another version next week and this can come in.

robbyxp1:
Now with an update from me on top, moved the enum to the correct place next to the function which uses it, and moved the text up to route, as text should be avoided as much as possible in the elitedangerous helper functions.

Thanks for the push, will be in 11.1.2 this week.

PR #2648, opened 2019-11-03
Fix some cases of excessively long DB connections

klightspeed:
* Fix the EDSM log fetch holding the database connection for an excessively long time
* Move the journal log reader out of the DB connection
* Pass the DB connection to tlu.Get in order to eliminate one cause of DB re-entrancy
* Log the DB type in logged re-entrancy / hold time messages


robbyxp1:
Looks good, many thanks.

PR #2665, opened 2019-12-08
EDCommander.cs updates

Elite-IGAU:
Fixed error in previous PR ( line 140 of EDCommander.cs )


robbyxp1:
I've further tidied up the class - its one of the oldest so needed more tlc.  I've also corrected an error on update:                         cmd.Parameters["@SyncToIGAU"].Value = edcmdr.SyncToIGAU;

rob

PR #2666, opened 2019-12-09
Fix ConfigurableForm DialogResult / Close race

### LINKED ISSUE #2663: Issues when using mono on linux
Firstly, the "Settings" tab flickers violently for a good 1-5mins after program startup, eventually this stops and i'm able to edit settings normally from that point forward, until then, no elements on that tab are displayed, just the outline boxes of different areas that flicker in/out of existance rapidly.

Secondly, setting the "minimum scan value" doesn't work. It always immediately reverts back to the default "500000" once I hit OK on the popup, no matter what number I enter.

There's no console output in a terminal, and I'm not sure where the programs logs (if any) are stored so..

klightspeed:
This should fix #2663

Depends on EDDiscovery/ExtendedControls#16, EDDiscovery/ActionLanguage#1, and EDDiscovery/BaseUtilities#16

### DIFF
--- a/EDDiscovery/3DMap/FormMap.cs
+++ b/EDDiscovery/3DMap/FormMap.cs
@@ -2356,8 +2356,7 @@ namespace EDDiscovery
             {
                 if (controlname == "OK" || controlname == "Cancel")
                 {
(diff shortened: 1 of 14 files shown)


robbyxp1:
Bit busy to bring in right now.. will have a look later in the week and bring in. Thanks

PR #2673, opened 2019-12-18
Compare system name and StarSystem case insensitively

klightspeed:
The system names on EDSM have inconsistent casing for some systems.
Compare the system names case insensitively when assigning bodies to systems.


robbyxp1:
thanks

PR #2675, opened 2019-12-18
Don't re-enter PlayJournalList loop

klightspeed:
This should prevent journal events from being processed in reverse order when event processing exceeds the delay timer interval.

This should fix the cause of the out-of-order events discovered in EDSM-NET/FrontEnd#331

### DIFF
--- a/EDDiscovery/EDDiscoveryControllerNewEntry.cs
+++ b/EDDiscovery/EDDiscoveryControllerNewEntry.cs
@@ -26,6 +26,7 @@ namespace EDDiscovery
     {
         private Queue<JournalEntry> journalqueue = new Queue<JournalEntry>();
         private System.Threading.Timer journalqueuedelaytimer;
+        private bool journalqueuedelaytimersemaphore;
 
         public void NewEntry(JournalEntry je)        // on UI thread. hooked into journal monitor and receives new entries.. Also call if you programatically add an entry
         {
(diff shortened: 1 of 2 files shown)


robbyxp1:
I'm trying to understand.  There may be something fundamentally wrong.

What your saying is that the timer tick, causing delay play, is causing a double play.

Since they are all on the UI thread, and if we are in the middle of PlayJournalList due to a timer from DelayPlay, then another NewEntry could cause PlayJournalList to be called? 

They are all supposed to be on the UI thread, how do we somehow get into PlayJournalList again..

robbyxp1:
Same comment the other way, if we are PlayJournalList() with a set of stuff, and the timer times out, it queues up PlayJournalList() Async.  Would that not wait till the executing PlayJournalList finishes?

klightspeed:
`Application.DoEvents()` runs the message pump, which includes running any pending `Control.Invoke` / `Control.InvokeAsync` calls.

robbyxp1:
Well, it seems like we have not got to the bottom of it - it could be a problem with the reader, or the issuing of call to NewEntry, as well as the idea that the timer is causing reentrancy.

i'll try and think of a way to reproduce.

klightspeed:
It would be great if we could get a stack trace of when this occurs - perhaps add a warning (and log the full stack trace) near the end of `Controller_NewEntrySecond` when e.g. the timestamp goes backwards or when an entry is not the last entry in the history?

The fact that the entries have the correct system info filled suggests that something it causing new events to be processed before the current event has been sent to EDDN but after it has been added to the history list.

Only Non-Queued events should be processed while processing a `SendMessage` in a control, while `Invoke` and `BeginInvoke` both use queued messages, which should only be processed either in the main message pump (i.e. `Application.Run`) or in the message pump of a modal dialog (e.g. `Message.Show`) or dropdown.

klightspeed:
Either a lot of apps (all of them being desktop applications, and notably excluding EDMC and EDSM) deliver events out of order under certain circumstances, or EDDN is sometimes delivering those events out of order.

Affected apps (since December 2018):
EDDiscovery: 5974 out-of-order scans from 1002 unique uploaderIDs on 148 days
EliteLogAgent: 30277 out-of-order scans from 10411 unique uploaderIDs on 334 days
EDDI: 817 out-of-order scans from 626 unique uploaderIDs on 221 days
Elite G19s Companion App: 15 out-of-order scans from 12 unique uploaderIDs on 7 days
Moonlight: 14 out-of-order scans from 11 unique uploaderIDs on 10 days
EDWfSvc: 4 out-of-order scans from 1 unique uploaderID on 1 day
E:D Market Connector [Linux]: 2 out-of-order scans from 2 unique uploaderIDs on 2 days
EDAOS: 1 out-of-order scan

robbyxp1:
So i've just written blamtest in eddtest, and blamed at 200ms intervals lots of FSD events, exercising the system, and checking in sendEDDN that they are in order. They are.

So with no absolute evidence its our fault, lets just put your debugging statement into eddiscoveryform.cs and issue the code, and see if anyone sees the error..

thanks

robbyxp1:
So i'm presuming, except for the debug bit, that this is now not needed and we fix in another way.

robbyxp1:
Closing this, since i think we fixed it another way. Thanks for the investigation

PR #2677, opened 2019-12-19
Updated PR

Elite-IGAU:
Had to add Extraversion.cs and comment out a bunch of unused code - now cleanly compiles. 

A few questions. 

Is IGAUSync.cs even necessary? Seems like alot of that code did some heavy lifting for EGO. 

Seems like for a simple HTTPS POST shipper (if the option is enabled in settings) IGAUClass.cs can do all the functions?

From the previous PR - all that needs to be done is something similar to what is being done in ELITE Observatory:

public static void ProcessCodex(CodexEntry codexEntry)

        {
                  if (Properties.Observatory.Default.SendToIGAU)
                  {
                  JObject POST_content = JObject.FromObject(
                new {
                    timestamp = codexEntry.Timestamp.ToString("yyyy-MM-ddTHH:mm:ssZ"),
                    Name_Localised = codexEntry.NameLocalised,
                    codexEntry.System
                });
                  var request = new System.Net.Http.HttpRequestMessage
                   {
                        Method = HttpMethod.Post,
                        RequestUri = new Uri($"https://ddss70885k.execute-api.us-west-1.amazonaws.com/Prod"),
                        Content = new StringContent(POST_content.ToString(Formatting.None))
                   };
                   HttpClient.SendRequest(request);
                  }
        }

### DIFF
--- a/EliteDangerous/EliteDangerous.csproj
+++ b/EliteDangerous/EliteDangerous.csproj
@@ -93,6 +93,8 @@
     <Compile Include="EliteDangerous\EDCalculations.cs" />
     <Compile Include="DB\EDCommander.cs" />
     <Compile Include="EliteDangerous\Translations.cs" />
+    <Compile Include="IGAU\IGAUClass.cs" />
+    <Compile Include="IGAU\IGAUSync.cs" />
     <Compile Include="Journal\EDJournalClass.cs" />
     <Compile Include="Journal\EDJournalReader.cs" />
     <Compile Include="Journal\EDJournalMonitorWatcher.cs" />
--- a/EliteDangerous/IGAU/IGAUClass.cs
+++ b/EliteDangerous/IGAU/IGAUClass.cs
@@ -1,0 +1,77 @@
+﻿/*
+ * Copyright © 2016 EDDiscovery development team
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not use this
+ * file except in compliance with the License. You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software distributed under
+ * the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+ * ANY KIND, either express or implied. See the License for the specific language
+ * governing permissions and limitations under the License.
+ *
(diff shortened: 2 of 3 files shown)


klightspeed:
The separate sync thread is to ensure that any delays or errors in posting don't either hang the UI or cause an unhandled exception.  It is there also for if you need to rate-limit the data being posted, or retry any failed requests.

Elite-IGAU:
Gotcha.  If I'm understanding right then, I need to put the HTTPS POST output code in IGAUSync.cs then.  Haven't done threaded apps before.

klightspeed:
The way we've done it is that `xxxxSync.cs` handles the queueing / rate limiting / retrying, while `xxxxClass.cs` handles the actual requests.

robbyxp1:
so, i'm not going to bring this in, because i want to keep the tip free for updates. I plan to do another version soon. Carry on developing on this thread..

Elite-IGAU:
Right, - so how will we manage conflicts and such if the current DB entry the IGAU stuff is using? Maybe merge in the database additions?

Elite-IGAU:
Got it. To be honest, queuing/rate limiting is a bit beyond my skill - if you look at the IGAU's EDMC plugin, and our code in Elite Observatory. it's pretty dead simple code.  

Would it be a major issue/problem for you if xxxxSync.cs is a no-op and/or removed?

Also I think I understand what you meant by not committing - obviously these classes are not production ready. I don't understand what you men by "tip free for updates", but I get your overall message. 

IGAUClass.cs will have the actual HTTPS POST code shipping to API gateway. I did drop you a note on Discord asking about the potential use of global variables that EDDiscovery might make available to other processes. 

> The way we've done it is that `xxxxSync.cs` handles the queueing / rate limiting / retrying, while `xxxxClass.cs` handles the actual requests.

klightspeed:
> Got it. To be honest, queuing/rate limiting is a bit beyond my skill - if you look at the IGAU's EDMC plugin, and our code in Elite Observatory. it's pretty dead simple code.

The EDMC plugin is dead simple because EDMC is dead simple, and doesn't need to worry about trying to display data in real-time.

Elite-IGAU:
Got it. At this point I'll close this PR and let it die.  Will need to get some time with you on Discord and work out how you want things rate limited, so there's no chance of the code breaking the app. Thanks again for the quick, and insightful replies.

PR #2680, opened 2019-12-21
Hide filter form if already shown

### LINKED ISSUE #2678: can't filter history
### Description
Clicking the history filter.

klightspeed:
This should fix #2678


robbyxp1:
ta

PR #2682, opened 2019-12-22
Quote urlacl username

### LINKED ISSUE #2681: Web server doesn't start (possibly for Windows usernames with spaces in it)
I attempted to start the web server using the instructions here:
https://github.com/EDDiscovery/EDDiscovery/wiki/Web-Server#turning-on-web-server

After checking the box, I hit "yes" and granted Administrator privileges.
However, after Powershell finished, I received "Did not start - click OK to configure windows" and the checkbox remained unchecked (leaving the "Test" button unclickable.

I checked the Windows firewall and saw the expected firewall rule was in there.

I dug in the code a bit and found in UserControlSettings.cs that two separate PowerShell commands are run: one that sets the firewall rule and one for urlacls.

I found that manually running the second in PowerShell (as administrator) resolved the problem for me:
`netsh http add urlacl url = http://*:6502/ user="Devil Boy"`

I honestly can't tell what the command EDDiscovery attempts is, but I suspect that this may have occured because my Windows username has a space in it.

Please let me now if I can provide any further information.

klightspeed:
This should fix #2681


robbyxp1:
thanks

PR #2691, opened 2019-12-28
Add galmap and journal systems to system cache

klightspeed:
This should make controls that rely on the system cache at least partially usable when the full system database is unavailable.


robbyxp1:
Okay, i'm going to hold this feature upgrade for a while, maybe a week or so, until we see if the main branch needs further fixes..

robbyxp1:
Thanks.

PR #2722, opened 2020-02-21
Disable most visited chart under mono

klightspeed:
Mono only implements enough of `System.Windows.Forms.DataVisualization.Charting` to get stuff using it to compile.


robbyxp1:
Ta.

PR #2723, opened 2020-02-21
Exclude LocalMap when running under Mono

klightspeed:
LocalMap depends upon System.Windows.Forms.DataVizualization.Charting, which is a stub under Mono (most methods are not implemented).


robbyxp1:
And doesn't the other one which Spadino did too? I'm going to do a release in a mo so this may not get in..

klightspeed:
Plot uses OxyPlot.WindowsForms, which appears to do all of the drawing using standard GDI+ primitives (as exposed by `System.Windows.Forms` and `System.Drawing`).

PR #2727, opened 2020-02-22
Fix IndexOutOfRangeException in 36386f0

### LINKED ISSUE #2726: Sorting under Materials Panel causes unhandled UI exception
Under the Materials panel, about every 2-3 times I sort by one of the columns, I get a UI error.  I can hit continue and will keep working, then same thing again after a couple column sorts.  It does not matter which columns, I went ahead and randomly clicked them for a while to see if it kept happening.  You can still use it, just have to hit the continue button is all.  Can't see anything else on the user end of the experience.  Here is the diagnostic info.  Thanks for working on such an incredible piece of software and have a good one!

klightspeed:
Fixes: 36386f0ab3bec849e671c4aac3e7abdcbcfd5c69

This should fix #2726


robbyxp1:
Always catches me, thanks

Issue #2730, opened 2020-02-29
Exception on 29 Feb 3306


Ejsstiil:
### Description
App crash

### Exception Details:
>EDDiscovery v11.2.5.0 There was an unhandled UI exception.
>```
>==== BEGIN ====
>System.ArgumentOutOfRangeException: Year, Month, and Day parameters describe an un-representable DateTime.
>   at System.DateTime.DateToTicks(Int32 year, Int32 month, Int32 day)
>   at EDDiscovery.EDDConfig.ConvertTimeToSelectedFromUTC(DateTime t) in C:\Code\EDDiscovery\EDDiscovery\EDDConfig.cs:line 154
>   at EliteDangerousCore.JournalEvents.JournalMissionAccepted.MissionBasicInfo() in C:\Code\EDDiscovery\EliteDangerous\JournalEvents\JournalMissions.cs:line 208
>   at EliteDangerousCore.JournalEvents.JournalMissionAccepted.FillInformation(String& info, String& detailed) in C:\Code\EDDiscovery\EliteDangerous\JournalEvents\JournalMissions.cs:line 192
>   at EDDiscovery.UserControls.UserControlTravelGrid.CreateHistoryRow(HistoryEntry item, String search) in C:\Code\EDDiscovery\EDDiscovery\UserControls\History\UserControlTravelGrid.cs:line 446
>   at EDDiscovery.UserControls.UserControlTravelGrid.HistoryChanged(HistoryList hl, Boolean disablesorting) in C:\Code\EDDiscovery\EDDiscovery\UserControls\History\UserControlTravelGrid.cs:line 259
>   at System.Action`1.Invoke(T obj)
>   at EDDiscovery.EDDiscoveryController.ForegroundHistoryRefreshCompleteonUI(HistoryList hist) in C:\Code\EDDiscovery\EDDiscovery\EDDiscoveryControllerHistory.cs:line 192
>===== END =====
>```

I believe that this has something to do with 29 Feb date trouble

klightspeed:
Yeah - that weird leap year in 3302 and 3306, and the lack of a leap year in 3300 and 3304.

Perhaps it should subtract 86 (or 1286) years if the year is greater than or equal to 3300?

PR #2731, opened 2020-02-29
Fix exception on 29 Feb game time

klightspeed:
29 Feb 3306 is not representable as a DateTime

This should fix the crash in #2730, though 29 Feb will be converted to 01 Mar when converting to game time.

### DIFF
--- a/EDDiscovery/EDDConfig.cs
+++ b/EDDiscovery/EDDConfig.cs
@@ -151,7 +151,7 @@ namespace EDDiscovery
             {
                 t = t.AddYears(1286);  
                 if (!DateTimeInRangeForGame(t))
-                    t = new DateTime(DateTime.UtcNow.Year+1286, DateTime.UtcNow.Month, DateTime.UtcNow.Day);
+                    t = DateTime.UtcNow.AddYears(1286);
                 return t;
             }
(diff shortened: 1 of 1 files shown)


robbyxp1:
Doh for Frontier to pick a start year not synchronised with leap years.

i'll pull in this week.

Ejsstiil:
Seems that having Pilot Jornal files containing this date already causes main program to freeze. Now every start/restart of EDDiscovery produces huge freeze. Even game relog does the same.
Tried purge all (except game logs), even fresh install not helped.
I am using this without any Star Maps.

klightspeed:
@Ejsstiil do you still get the freezes with the build at <https://ci.appveyor.com/project/EDDiscovery/eddiscovery/builds/31144032/artifacts>?

Ejsstiil:
Tried and it still freezes... every ED relog/restart freezes EDD which was fine before 29 Feb.
I will try to remove ED Journals with 29 Feb entries and will test it more.

Ejsstiil:
> I will try to remove ED Journals with 29 Feb entries and will test it more.
Freezes are no related with this - confirmed. Its somewhere within /Data folder.

Please close this

PR #2734, opened 2020-03-06
Test IGAU Data shipping

Elite-IGAU:
I had to make a change to the .csproj file for the project to build and run. Thanks, StackExchange! 

Uncommented lines in IGAUSync.cs

Added a line to strip info from raw name field. 
Attempting to string.ToLower results in an error. (see commented line)

The build loads and runs fine, however when I right click in history and attempt to process a CodexEntry line, I get this debug error in the lower right hand pane:
`Exception Object reference not set to an instance of an object.
   at EliteDangerousCore.IGAU.IGAUClass.PostMessage(JObject msg, Boolean& recordSet) in \EDDiscovery\EliteDangerous\IGAU\IGAUClass.cs:line 84
   at EliteDangerousCore.IGAU.IGAUSync.SyncThread() in \EDDiscovery\EliteDangerous\IGAU\IGAUSync.cs:line 93`

### DIFF
--- a/EDDiscovery/EDDiscovery.csproj
+++ b/EDDiscovery/EDDiscovery.csproj
@@ -1134,10 +1134,8 @@
       </ZipFiles>
       <Zipfiles Include="$(OutDir)x64\SQLite.Interop.dll">
         <Name>x64\SQLite.Interop.dll</Name>
-          <CopyToOutputDirectory>Always</CopyToOutputDirectory>
       </Zipfiles>
       <Zipfiles Include="$(OutDir)x86\SQLite.Interop.dll">
-          <CopyToOutputDirectory>Always</CopyToOutputDirectory>
         <Name>x86\SQLite.Interop.dll</Name>
       </Zipfiles>
     </ItemGroup>
--- a/EDDiscovery/EDDiscovery.csproj
+++ b/EDDiscovery/EDDiscovery.csproj
@@ -1134,8 +1134,10 @@
       </ZipFiles>
       <Zipfiles Include="$(OutDir)x64\SQLite.Interop.dll">
         <Name>x64\SQLite.Interop.dll</Name>
+          <CopyToOutputDirectory>Always</CopyToOutputDirectory>
       </Zipfiles>
       <Zipfiles Include="$(OutDir)x86\SQLite.Interop.dll">
+          <CopyToOutputDirectory>Always</CopyToOutputDirectory>
         <Name>x86\SQLite.Interop.dll</Name>
       </Zipfiles>
     </ItemGroup>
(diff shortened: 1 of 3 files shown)


EDDiscovery/EDDiscovery.csproj
```diff
@@ -1134,8 +1134,10 @@
       </ZipFiles>
       <Zipfiles Include="$(OutDir)x64\SQLite.Interop.dll">
         <Name>x64\SQLite.Interop.dll</Name>
+          <CopyToOutputDirectory>Always</CopyToOutputDirectory>
```
klightspeed: Trying to work out how this would make a difference - you're saying to copy the file from the output directory to the output directory.

klightspeed:
> Exception Object reference not set to an instance of an object.
> at EliteDangerousCore.IGAU.IGAUClass.PostMessage(JObject msg, Boolean& recordSet) in \EDDiscovery\EliteDangerous\IGAU\IGAUClass.cs:line 84

If you put a breakpoint on that line, what is the response from the server?
```cs
                BaseUtils.ResponseData resp = RequestPost(msg.ToString(), "");

                JObject result = JObject.Parse(resp.Body);
                JObject res = (JObject)result["response"];
                if ((bool)res["is_valid"])   // <<<<<<
```

EDDiscovery/EDDiscovery.csproj
robbyxp1: I agree, won't pull that in.

robbyxp1:
Some queries need answering before we pull this in..

EDDiscovery/EDDiscovery.csproj
Elite-IGAU: I looked up the error I was getting when building, and saw this post.  Gave it a shot. Code built and ran afterwards. 

https://stackoverflow.com/questions/13028069/unable-to-load-dll-sqlite-interop-dll

Elite-IGAU:
Totally understand - I'm heading to bed, and will run again with a breakpoint in the AM.  Will let you know the server output. Any tips on resolving the issue with string.ToLower?

EDDiscovery/EDDiscovery.csproj
klightspeed: I suspect that modifying the csproj outside visual studio resulted in the project being rebuilt, thus pulling the interop files from the EliteDangerous project, which has:

https://github.com/EDDiscovery/EDDiscovery/blob/a595339420e569e31d18e2aa21fa5e9f9d299818/EliteDangerous/EliteDangerous.csproj#L17-L20

klightspeed:
> Any tips on resolving the issue with string.ToLower?

Same - set a breakpoint in the debugger.  You could also enable breaking on whatever exception is being thrown (Debug -> Windows -> Exception Settings :: Common Language Runtime Exceptions), but I don't know how many swallows have nested since logging first-chance exceptions was last disabled.

PR #2740, opened 2020-03-07
update IGAUSync.cs

Elite-IGAU:
Updated redundant (and broken) error check to a simple "transmitted" notification.  Aside from this minor change, it seems like all the IGAU code is working properly. 

Feel free to ignore the changes to the other two files. 

Every time I check out a fresh copy of the codebase, Visual Studio complains about ExtraVersion.cs not being present, so I have to manually create a blank file.


klightspeed:
Building once should create ExtraVersion.cs

Unfortunately including ExtraVersion.cs in the repo would cause it to no longer be ignored, and so every new commit would include a change to it.

klightspeed:
I have created #2741 which uses the return from PostMessage instead of the left-over out parameter from the copy-paste from EGO.

PR #2742, opened 2020-03-08
Added a switch to show/hide values in Surveyor

spadino:
This PR add a switch to Surveyor which allow to show/hide values.


robbyxp1:
Cheers

PR #2747, opened 2020-03-15
Add options to set low process priority

klightspeed:


robbyxp1:
See my comment in devs..

Also, should we make these tick boxes on settings?  EDDConfig items as well?

PR #2754, opened 2020-03-21
Fix NRE when ship is unknown

### LINKED ISSUE #2750: Unhandled UI Exception

klightspeed:
This should fix #2750


robbyxp1:
Okay, i pushed the SI check only in inarasync and found one other possible place in there.

thanks

PR #2757, opened 2020-03-28
Defer loading journal view until visible

iainross:
First 1k rows are always added to the grid so something will be there when brought into view, and events since initialisation are prioritised over populating the tail end of the view.

Will only impact journal views that are initialised in background tabs on start up, anything opened during a session will behave as before.


robbyxp1:
So, given i've solved this in another way, i'll discard. Thanks for the idea how to fix it and pointing it out.

PR #2765, opened 2020-04-13
Add journal events for Fleet Carriers

klightspeed:


robbyxp1:
Note they are adding more journal entries, as I said, suggest we wait until after 11.4 and when they issue a new journal manual.

PR #2770, opened 2020-04-19
Await on EDSM Bodies fetch to free up UI

robbyxp1:
A big one.  Now uses an async pattern to get EDSM body data using starscan Find system.
This means the UI is not frozen during these web lookups and is more fluid.
Since async functions can't lock, the only lock to prevent multiple web lookups of the same data (from multiple panels) is in the task.  The task locks and allows only 1 lookup, futher lookups is stalled in their tasks. The first task completes, further tasks will see the data in the dictionary caches and not lookup.
C# (research shows) frees up awaits in order, so the first one runs first, sees it was not in the cache, and adds the data.  The secoond on sees it was from the cache, and does not add it in.
This appears to be working - lots of debug is left in for now to test this.

Also cleaned up UserControlEstimatedValues to give it an EDSM on/off button and a column saying if the data is from EDSM.

### DIFF
--- a/EDDiscovery/Actions/ActionsEDDCmds/ActionScanStar.cs
+++ b/EDDiscovery/Actions/ActionsEDDCmds/ActionScanStar.cs
@@ -81,7 +81,7 @@ namespace EDDiscovery.Actions
                         sc.EDSMID = 0;
                     }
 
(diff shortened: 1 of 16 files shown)


robbyxp1:
Bring in and pray

PR #2775, opened 2020-04-24
Abandon missions that are missing from missions event

### LINKED ISSUE #2774: Failed mission are not removed
In the missions screen the missions that failed just stay there as active missions until the End Date.
You can't do anything with them anymore, they're no longer available in the game but you can't get rid of them.

klightspeed:
This should fix #2774

### DIFF
--- a/EliteDangerous/EliteDangerous/MissionList.cs
+++ b/EliteDangerous/EliteDangerous/MissionList.cs
@@ -77,7 +77,7 @@ namespace EliteDangerousCore
         {
             get
             {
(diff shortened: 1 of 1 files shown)


robbyxp1:
I'm unsure about this. Holding for now.

robbyxp1:
I'll do this in .2, want to do minimal changes for .1

robbyxp1:
This breaks the contract of the StateTypes, which states that JournalMissionCompleted is not null if state is Completed.

So it needs a little nerfing, even if its just a description. 

Bring in tomorrow.

PR #2782, opened 2020-04-27
Don't show distance from star for co-orbital planets

### LINKED ISSUE #2781: Ditance divided by 1000 for some bodies in scan module
EDDiscovery v11.4.1.0

In some systems, distances are incorectly reported for some bodies and seem to be divided by 1000
Here is an example for Swoilz RO-X d2-26 system where 6 bodies are incorrect. In the info box we can see the correct distance :
edd_distance_bug
The new habitable zone function seems to read the wrong distance too.

klightspeed:
Perhaps DistanceFromArrivalLS can be used if it's a single-star system, or if the planet is orbiting the arrival star.

For Co-orbital planets, the SemiMajorAxis is for the orbit around the barycentre, not around the star.

This should resolve #2781

### DIFF
--- a/EDDiscovery/UserControls/Helpers/ScanDisplayUserControl.cs
+++ b/EDDiscovery/UserControls/Helpers/ScanDisplayUserControl.cs
@@ -183,7 +183,7 @@ namespace EDDiscovery.UserControls
                             {
                                 List<ExtPictureBox.ImageElement> pc = new List<ExtPictureBox.ImageElement>();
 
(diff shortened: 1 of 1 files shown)


robbyxp1:
thanks. I'm working on a barycentre thingy so will incorporate that into it.

robbyxp1:
And yes, lets use distance from Arrival for co-orbiting.. i'll work it all into a push

robbyxp1:
Incorporated into my upcoming barycentre push, thanks

PR #2792, opened 2020-05-01
Fix barycentre bar for planets around stellar barycentre

klightspeed:


robbyxp1:
Going to think about this because its coding in some assumptions about bodyid ordering.

klightspeed:
Barycentres always have a lower body ID than any of their children, and top-level stars always come before any planetary bodies.
image
Think of trees - the stars and their barycentres are nodes which are visited in order (i.e. the parent is visited before its children), and then those are visited depth-first when enumerating planetary bodies (i.e. the children are visited before the parent), with the bodies therein being visited in order (i.e. the parent is visited before its children).

klightspeed:
Without the fix:
image

With the fix:
image

robbyxp1:
Closing, see comment in discord.  This was showing up a fundamental problem with the drawing algo.

PR #2793, opened 2020-05-01
Improve performance of population of potentially large grids

iainross:
Populates three DataGridView controls that can potentially have a lot of rows by making a list of rows outside the grid and adding as a range to the control rather than adding individually to the control for huge performance benefits when the control is not visible.


robbyxp1:
Thanks

PR #2794, opened 2020-05-02
Fix grids skipping 1500 entries after the first 500

klightspeed:
The post part of a for loop is executed after the last statement in the for loop.  As a result, the loop was processing 500 entries, then incrementing the index by 2000.


robbyxp1:
One of my more dumber bits of code, thanks

PR #2796, opened 2020-05-04
Feature/hd icons

spadino:
## Icon set revamp

Preliminary work:

- Replace all icons with high definition 512px images
- Add many variants for many body classes
- Moves bodies icons to Bodies/{Stars,Giants,Planets}, for better organization of variants
- Refactor JournalScan.cs where icons are associated to body classes, for a cleaner logic


robbyxp1:
Thanks. This will need to come in after the release for this week.

PR #2798, opened 2020-05-08
Revamp of the IconSet;

spadino:
Improve all body classes images to high definition, 512px icons. It also add many variations for most classes.


robbyxp1:
So, I want this to be a key feature of 11.5 (got to give them a reason to update) so i'm holding off until we have the final 11.4..

spadino:
Will push a cleaner one

PR #2799, opened 2020-05-09
Move EDSM journal discard list fetch to background

klightspeed:
This should fix an issue where a slow response from EDSM will cause the UI to freeze when Sync to EDSM is enabled.

### DIFF
--- a/EliteDangerous/EDSM/EDSMJournalSync.cs
+++ b/EliteDangerous/EDSM/EDSMJournalSync.cs
@@ -167,7 +167,7 @@ namespace EliteDangerousCore.EDSM
 
         // Called by Perform, by Sync, by above.
 
-        public static bool SendEDSMEvents(Action<string> log, IEnumerable<HistoryEntry> helist, bool manual = false)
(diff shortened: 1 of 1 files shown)


robbyxp1:
Good idea.

robbyxp1:
Okay, but I have another suggestion. Why not just do the whole discard check, with time out, around here:

                        historyevent.Reset();
                        Action<string> logger = hqe.Logger;

                        List<HistoryEntry> hl = new List<HistoryEntry>() { first };

                        string logline = $"Adding {first.EntryType.ToString()} event to EDSM journal sync ({first.EventSummary})";
                        System.Diagnostics.Trace.WriteLine(logline);
                        //removed too verbose if (!manual)  hqe.Logger?.Invoke(logline);


Say just before making the list.. then its in the sync thread, does not change EDControllerMain...

klightspeed:
I have now moved the reject list update into the EDSM sync thread, and added a check after dequeuing the event to check if it's on the reject list.

robbyxp1:
Thanks.

PR #2800, opened 2020-05-09
Reduce time spent updating Stats panel

iainross:
Heat warning/damage, fuel scooping and landing count have been removed from the Travel tab.  This, combined with new helper methods on `HistoryList` to avoid enumerating the same journal type list repeatedly, brings the refresh time to well under a second.  

Total first discoveries removed from General tab as it's no longer a meaningful stat with the lack of a count on selling pages now.

Removal of those journals from essential stats events reduced the number of journals needed in memory significantly (312,981 to 272,537 on my dataset with only stats entries older than 7d) for a general performance bump.

All tabs now refresh only on relevant journal entries.

### DIFF
--- a/EDDiscovery/UserControls/CurrentState/UserControlStats.cs
+++ b/EDDiscovery/UserControls/CurrentState/UserControlStats.cs
@@ -29,9 +29,15 @@ namespace EDDiscovery.UserControls
 {
     public partial class UserControlStats : UserControlCommonBase
     {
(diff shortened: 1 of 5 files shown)


robbyxp1:
Thanks. 

Have you got some sort of global autocorrect on for spaces, because there was a lot of whitespace corrections..

iainross:
Yeah, been using the VS on my own desktop for work stuff with the whole lockdown thing and I need to have the Power Tools `Format On Save` option turned on for that.

PR #2801, opened 2020-05-10
Icon set revamp

spadino:
Clean revamping of default icons. Provide high definition icons for all body classes, as well as many variations.


robbyxp1:
Thanks

Put StarClassID back to the original name StarTypeID please.  The Action system enumerates the names of fields and properties and by changing the name you'll be making the voice action pack fail.

In general, you can't change the name of journal properties/fields without double checking they have not been used by the Action packs. 

It will also reduce the changes

Once you've done this, i'll check it.

robbyxp1:
Re above about  StarClassID

spadino:
Sure, no problem!

PR #2810, opened 2020-05-17
Fix duplication in Market Data Panel

### LINKED ISSUE #2805: BUG Market Data
When carrying the relevant commodity. VO and LTD cargo show on a separate additional line such that you have 2 entries for these commodities, one showing the market data and 0 cargo and one showing no market data and the correct cargo.

klightspeed:
This should fix the duplication of commodities in the market data panel when the symbol does not match the english name.  Examples are Low Temperature Diamonds (symbol lowtemperaturediamond) and Viod Opal (symbol opal).

This should fix #2805


robbyxp1:
Ta. Was going to investigate this later but you beat me to it.

PR #2818, opened 2020-05-27
Visited Stars Added Scan Value to Excel Export

carcher99:
Visited Stars - Added Scan Value to Excel Export to match what is seen onscreen


robbyxp1:
Thanks

PR #2820, opened 2020-05-27
Fix up carriers

robbyxp1:
Debugged against journal input


EliteDangerous/JournalEvents/JournalNavRoute.cs
```diff
@@ -37,7 +37,7 @@ private void Rescan(JObject evt)
 
                 foreach (JObject jo in route)
                 {
-                    var starsys = jo["StarSystem"];
+                    var starsys = jo["StarSystem"];         // StarSystem was not in initial beta of this
```
klightspeed, line 40: StarSystem was the system address in the initial beta

PR #2822, opened 2020-05-27
Feature/hdicons fixed

spadino:


robbyxp1:
Hmm

SizeLarge icon is shown to be deleted.. why?

Otherwise ok..

spadino:
Oh, don't know why. It should be there...

PR #2825, opened 2020-05-27
Adding Travel distance to Visited Stars

carcher99:
Added Travel Distance from Arrival to the information column on the Visited Stars page


robbyxp1:
thanks

PR #2827, opened 2020-05-29
Prevent runtime error when calling GetLastHistoryEntry with a null frominclusive parameter

iainross:
Noticed this could happen from an Outfitting panel, protected generally (and removed the now redundant protection from the stats panel)

Format on save continuing to apply work code standards for spaces... :)


robbyxp1:
Spaces..Tssks.. ;-)

robbyxp1:
thanks

Issue #2345, opened 2019-01-18
Feature request - Mapped column in Estimated Vaules grid


Kingdud:
The Estimated Values 'pane' is very useful and compact. That said, given the 3.3 exploration changes, I'd really love if I could tweak two things about it:

1. Add a third column to the window which is a simple checkbox / x that will tells me if I have already mapped a given planet or not. If you'd like to re-use the green-globe asset from the Scans pane, that's fine too. Just some way for me to visually see, in that list, which systems I have mapped and which ones I haven't.
2. A way to sort the list in a 2-depth manner. IE: Primary sort by what is/isn't scanned yet, and secondary sort by system value (that way the 20 icy bodies stay at the bottom, and the higher value stuff comes to the top).

I don't know how hard either of those are, but I can say #1 is the one I want *far* more than #2. I'm hoping this is a simple change, but I don't know that for a fact.

robbyxp1:
We will look. Moving to projects.

PR #2828, opened 2020-05-31
#2345 mapped column in estimated values

carcher99:
Added Body Type and Mapped columns to the Estimated Values panel. Response to the feature request #2345 in the Projects - Future Ideas page

### DIFF
--- a/EDDiscovery/UserControls/ScansStars/UserControlEstimatedValues.Designer.cs
+++ b/EDDiscovery/UserControls/ScansStars/UserControlEstimatedValues.Designer.cs
@@ -155,6 +155,7 @@ namespace EDDiscovery.UserControls
             this.Mapped.HeaderText = "Mapped";
             this.Mapped.Name = "Mapped";
             this.Mapped.Width = 52;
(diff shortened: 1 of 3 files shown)


robbyxp1:
Okay, but if we are going to do this, why not also have a column showing if the body has been discovered?

carcher99:
Sure I can add that

carcher99:
..but thinking about it - if it is in estimated values list surely that implies it's discovered...what am i missing?

klightspeed:
> ..but thinking about it - if it is in estimated values list surely that implies it's discovered...what am i missing?

I think Robby means whether `WasDiscovered` is true on the scan, meaning that the body has been "tagged" by somebody else.

carcher99:
> I think Robby means whether `WasDiscovered` is true on the scan, meaning that the body has been "tagged" by somebody else.

Thanks - I'll get that done.

robbyxp1:
Yes, WasDiscovered, WasMapped..

carcher99:
Done - Obviously open to feedback gents.

robbyxp1:
thanks

PR #2830, opened 2020-06-06
Fixed Sorting in the Statistics Scan Panel

carcher99:
Fixed Sorting in the Statistics Scan Panel


robbyxp1:
Thanks, its going to be too late for this next release, but i'll look at it later this week/next.

robbyxp1:
Thanks, i've sorted it (!) another way by fixing the sortcompare function - it was calling an incorrect sorter for two tabs.

PR #2851, opened 2020-06-13
[11.5] Don't process NavRoute file if the game is spamming it

klightspeed:
This should work around the Elite Dangerous `NavRoute` spamming bug, where `NavRoute` events are spammed to the journal by the game.


robbyxp1:
Ta

robbyxp1:
Need to do this on Master?

klightspeed:
> Need to do this on Master?

This is in EDDiscovery/EliteDangerousCore#1

PR #2854, opened 2020-06-14
[11.5] #2853 Ignore events when in CQC

### LINKED ISSUE #2853: EDDiscovery records CQC Systems
I played CQC and EDDiscovery recorded the systems the match was played in.

klightspeed:
The only indication that we're in CQC is the music track, so use that to detect when we're in CQC.
This should fix #2853


robbyxp1:
Has conflicts, will fix later if you don't.

klightspeed:
Conflict resolved

PR #2858, opened 2020-06-21
Astroplot localmap

spadino:
Preliminary work on LocalMap to switch to AstroPlot; working, but need refinements. This PR is not intended to be incorporated, I just pushed to check if it can be merged with no troubles...


robbyxp1:
Note the errors

klightspeed:
`System.Management.Automation` (i.e. Powershell) is not available under Mono.  Is there a reason it would be bringing in Powershell?

spadino:
Yeah, many namespaces were added by their own by vstudio... 
That's ok, it compiled successfully in AppVeyor, so it really should because of those extra namespace added. Thanks, guys!

PR #2868, opened 2020-07-04
Replace LocalMap with AstroPlot

spadino:
Replace the broken LocalMap with a working yet basic implementation of the brand new AstroPlot. Many features yet to coming, but it already give something that wasn't available at the moment: a working 3D representation of neighborhood systems.

---

- Need to fix the tooltip threading issue
- Need improvements to the UI
- Planned refactors to avoid some code duplication
- ... ;)


robbyxp1:
This is broke. I think the best thing is if i bring in the baseutils and extended control updates, then you update your branch with those latest updated versions and repush this up

spadino:
Think so! Thanks

robbyxp1:
Rework with Ext controls and baseutils

PR #2874, opened 2020-07-13
Update EDDIcons with a new landable overlay for ScanPanel

spadino:
- It replace the landable overlay with a more refined, hand crafted one
- Replace three gas giants icons


robbyxp1:
Again, you've given me new images, removed some images, and not given a push for Journalscan.cs

GG1v11 is referenced by journalscan, so removing it breaks the code.

Try again ;-)

spadino:
Complemented by PR #9 in EliteDangerousCore submodule

PR #2875, opened 2020-07-14
Icons/update

spadino:
- Add some new icons, replace others
- Add a new landable overlay, hand crafted


robbyxp1:
deleting icons referenced by journalscan.

Please rework.

PR #2889, opened 2020-07-27
Website20

tecnosegugio:
This is for a better website for EDDiscovery.
I've used this modified version for several days and haven't found errors or missing images so now I'm trying to make my first "pull-request" on GitHub. ;-)
As I've said before, I'm not using the icons from the DLL of EDDiscovery but I'm using a directory "icons" under EDD. In this way I can change them easily.
Let me know what you think about this new website.


robbyxp1:
I appreciate the effort involved, but i'm not going to get into changing the website at this point.  I don't have the time or bandwidth to support that or accept in the push.

We will have to think about how to improve the website much later in the year.

robbyxp1:
Although i've closed the PR (do not leave PR pending) i've added it to the projects future ideas list so it won't be forgotten.

tecnosegugio:
OK, in the meantime I'll try to furher improve the website.
BTW: Thank you for your work on EDDiscovery ;-)

PR #2892, opened 2020-07-29
Fix case of GGHv7 and GGHv8

klightspeed:
These files where changed in the csproj, but not renamed using git mv, resulting in the Linux build failing.

Fixes: 40d2358 Include two missing icons in project


robbyxp1:
Hmmm i thought i'd fixed that..  I noticed that before. Must have been lost.

PR #2895, opened 2020-08-01
Update key used in ByShip tab in Stats panel

carcher99:
Using ShipKey ensures ships with same Id dont overwrite each other - In certain cases 2 ships can have same Id, and overwrite each other in the view.


klightspeed:
Given that ShipKey is constructed using ship type and ID, and the first column consists of ship type and ID, this would have no effect.

carcher99:
Not sure exactly the scenario - But I think when you have sold a ship, and then rebought the same ship type

In the shipInformationList class it updates the Id of a sold ship to 30000 

So I end up with the following when building the data for the panel DGV
First Key is asp:3, Value is 3:"Asp Explorer":"Asp":24
Second Key is asp:30000, Value is 3:"Asp Explorer":"Asp":22..

In the history list however all entries are listed against asp:3 not asp:30000 - End result is the stats for the second entry overwrite the first if you use Value.. If you use Key you get 2 rows. It does make a difference, for my data at least. 

Chris

PR #2904, opened 2020-08-12
Fix NRE when BeltData is null

### LINKED ISSUE #2903: No idea what caused this to pop up!

klightspeed:
This should fix #2903


robbyxp1:
thanks

PR #2932, opened 2020-10-09
Prefix username with user domain

### LINKED ISSUE #2931: Web server will not start
I have tried to enable the webserver but no longer have success.  I had it running once.   I enabled the option but it complains it cannot start.  I deleted the auto created FW rule and ran netsh http delete urlacl url=http://*:6502/ and tried to enable it again but it still will not start.

image
image

klightspeed:
This should fix #2931

### DIFF
--- a/EDDiscovery/UserControls/Settings/UserControlSettings.cs
+++ b/EDDiscovery/UserControls/Settings/UserControlSettings.cs
@@ -622,7 +622,7 @@ namespace EDDiscovery.UserControls
                                  "-Protocol TCP;" +
                                  "netsh http add urlacl " +
                                 $"url = http://*:{EDDConfig.Instance.WebServerPort.ToStringInvariant()}/ " +
-                                $"user=\"{Environment.GetEnvironmentVariable("USERNAME")}\"";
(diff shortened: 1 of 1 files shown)


robbyxp1:
I'm presuming this has been checked and worked on non domain computers?

klightspeed:
On local accounts (regardless of whether the computer is domain-joined or not), `%USERDOMAIN%` is the computer's machine name.

For example, my computer's name is `COPERNICUS`, and my username is `klightspeed` so it runs:
```
Powershell.exe -Command new-netfirewallrule -Name EDDiscovery -DisplayName EDDiscovery -Description Webserver -Direction Inbound -Action Allow -LocalPort 6502 -Protocol TCP;netsh http add urlacl url = http://*:6502/ user="COPERNICUS\klightspeed"
```

This works correctly on my computer (which I recently re-installed Windows on, so it didn't already have the firewall rule or URL ACL.

Note that the username and domain are still that for the local account, even if you sign onto Windows using a Microsoft account.

robbyxp1:
Cheers, don't have time at mo to test, so just wanted to make sure someone did.

PR #2947, opened 2020-11-10
Provide option to reset DB path if DB path not found

klightspeed:


robbyxp1:
Good idea, but don't you think it would be better placed in EDDiscoveryControllerMain, at 173..

EDDOptions should not really be active in this way..

robbyxp1:
Thanks.

I think if they don't reset, we should stop the program. I'll edit it a bit more.

PR #2949, opened 2020-11-17
Add Mission Accounting panel

keathmilligan:
This PR adds a new "Mission Accounting" panel that tabulates mission rewards - reputation gains, influence gains, credits and other rewards and displays them by faction. This feature is especially targeted at players participating in BGS activities. Like the missions panel, the display is filterable by date range.


robbyxp1:
Looks good, i'll bring it in v.soon.

robbyxp1:
Some comments coming up

EDDiscovery/UserControls/CurrentState/UserControlMissionAccounting.cs
```diff
@@ -0,0 +1,335 @@
...
+
+                        if (DateTime.Compare(ms.Completed.EventTimeUTC, startdateutc) >= 0 &&
+                            DateTime.Compare(ms.Completed.EventTimeUTC, enddateutc) <= 0)
+                        {
+                            System.Diagnostics.Debug.WriteLine(ms.Mission.Faction + " " + ms.Mission.Name + " " + ms.Completed.EventTimeUTC);
+                            total++;
+                            var faction = ms.Mission.Faction;
+                            int inf = 0;
+                            int rep = 0;
+                            foreach (var fe in ms.Completed.FactionEffects)
```
robbyxp1, line 201: Crashed due to no factionEffects on one of my missions, needs a null check

EDDiscovery/UserControls/CurrentState/UserControlMissionAccounting.cs
```diff
@@ -0,0 +1,335 @@
...
+                                if (fe.Faction == faction)
+                                {
+                                    if (fe.ReputationTrend == "UpGood" && fe.Reputation?.Length > 0)
+                                    {
+                                        rep = fe.Reputation.Length;
+                                    }
+
+                                    foreach (var si in fe.Influence)
+                                    {
+                                        //if (si.SystemAddress > 0)
```
robbyxp1, line 212: Need to get rid of commented out code..

robbyxp1 (commented):
Looks good otherwise.

We could do double click (and a right click option) Show all missions for faction, then it pops out a new window (like in UsercontrolTravelGrid line 768) with an InfoForm to show all the missions against date.. (or maybe a form with a DGV)... but if you review and fix those below i'll bring it in.

thanks for the push

keathmilligan:
Thanks, I'll add the pop-out in another push.

robbyxp1:
Thanks.  

Actually, the pop out, would probably share a lot of code with the missions panel grid - i would make a new MissionListUserControl in helpers and use that in both Missions and in the pop out.  This would prevent code duplication.

I'll bring this in, and we can consider the pop out later.

robbyxp1:
I've split out a MissionListUserControl, modified usercontrolmissions to use it.

I may add the info form later (which can be an ConfigurableForm so we don't have to create yet another Form).

robbyxp1:
Actually i'm coding it now, i want to get another EDD out soon.

robbyxp1:
Just to let you know, i'm going to rename this to the faction panel, and will be adding more faction information, such as trading stats, pvp kill stats, etc to it in the future.

thanks for the push

keathmilligan:
That sounds awesome. Looking forward to seeing it.

PR #2952, opened 2020-11-24
Add systems detail panel to UserControlFactions

keathmilligan:
Allow user to click on a faction row in the UserFactionsControl panel and display the systems involved in missions undertaken for that faction with the amount of influence gained in each system along with other basic information about the system.

### DIFF
--- a/EDDiscovery/UserControls/CurrentState/UserControlFactions.Designer.cs
+++ b/EDDiscovery/UserControls/CurrentState/UserControlFactions.Designer.cs
@@ -34,21 +34,38 @@
             this.labelTo = new System.Windows.Forms.Label();
             this.endDateTime = new ExtendedControls.ExtDateTimePicker();
             this.labelValue = new System.Windows.Forms.Label();
-            this.dataViewScrollerPanelFactions = new ExtendedControls.ExtPanelDataGridViewScroll();
             this.scrollBarFactions = new ExtendedControls.ExtScrollBar();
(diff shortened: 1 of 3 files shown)


keathmilligan:
User-experience wise, these changes are a bit at odds with the double-click functionality. A single-click displays the system detail in the split panel, while a double-click displays the mission detail in a pop-up. I was thinking one option might be to add tabs to the lower split panel to select between missions detail, systems detail and potentially other info in additional tabs later on.

robbyxp1:
Hmmm its moved on a bit from that, so i'll have to look later to see how this can be integrated.

keathmilligan:
Sounds good. Let me know what I can do to help.

robbyxp1:
Thanks. I've been added more stats to the program, and pushing them up to the factions panel.  

I think it will be best to have right click selects for mission list details, trade details, etc.. it would make more logical sense..

robbyxp1:
Hmm, your asking for information from EDSM for system data, in a large group, across all systems.

I've been explicitly asked not to bombard EDSM with too many queries, and Anthor also has a limiter on queries over a period of time to stop him being overloaded.  So I don't think this idea works. The other panels which ask info from EDSM tend to by design drip feed the info into the panel, or not have to ask too often.

I'll get the tip back together with my changes over the next few days and we can decide.

btw if you use discord, we have a server, which may be easier to talk together on.

robbyxp1:
I'm going to close this as it now conflicting.. Lets talk on discord.

PR #2966, opened 2020-12-20
Fix build error in tests

klightspeed:


robbyxp1:
Do we actually want to carry on maintaining these..?

PR #2970, opened 2020-12-22
Fix compile error in tests

klightspeed:
Fixes 9db6adf


robbyxp1:
Sorry, i never check them, thanks

PR #2984, opened 2021-01-10
Added extended BodyScanStatus

Keen13:
I find it useful for exploration to display a bit more information on the Scan panel about planet scan status.  My idea is to show it through various icons in place of current 'mapped' icon.

My code introduces the next scan info statuses:
        KnownFromEDSM,
        Scanned,
        ScannedAndWasMappedBySomeone,
        ScannedAndNotMappedByAnyone,
        MappedByYou
The last one corresponds to the existing 'mapped' icon.  I added icons for KnownFromEDSM (meaning that a body wasn't scanned by you in any way) and ScannedAndNotMappedByAnyone (meaning that you can get the 'first mapped' bonus for it).  No special icon is displayed for other statuses, though they can be added in future. See the screenshots: standard Mapped icon, KnownFromEDSM, ScannedAndNotMappedByAnyone.

image
image
image

Notes:
1) This feature also has changes in EliteDangerousCore. See https://github.com/EDDiscovery/EliteDangerousCore/pull/32
2) I don't think that the icon for 'KnownFromEDSM' status is good, I just took the same "EDSM" icon in different color. It would be nice to find something that looks clearer, especially for moons.
        
Please, take a look and see if it makes sense. I've been using this feature for a few weeks already and I find it quite useful.


robbyxp1:
Okay, The only problem I will have with this is how to explain it! We will need to update the Wiki, and the video i did a few months ago will be out of date!

I've been asked what those icons need about 1000 times!

Thanks, i'll bring it in when i'm ready to update those bits as well.

Keen13:
Thanks. Any ideas about better icon(s)?

robbyxp1:
On further consideration:

EDSM is already marked by a underscored name.   We don't want an icon for that, it will just clutter.

We can distinguish between mapped by you or mapped by another.  I think i'll take the idea and run with that.

So i'm afraid its not going to be pulled in.

PR #3106, opened 2021-06-14
Bring OnNewStarList up to main form

### LINKED ISSUE #3105: Add to... works on the main history tab but not on other history grid tabs.
Title says it all.
If (single or multiple) entries are selected on the main history tab the "Add to..." context menu options work just fine, but the same thing doesn't do anything if working with a seperate history panel.

klightspeed:
This should fix #3105

### DIFF
--- a/EDDiscovery/EDDiscoveryForm.cs
+++ b/EDDiscovery/EDDiscoveryForm.cs
@@ -15,6 +15,7 @@
  */
 
 using EDDiscovery.Forms;
+using EDDiscovery.UserControls;
 using EliteDangerousCore;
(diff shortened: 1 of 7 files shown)


robbyxp1:
No sure about this, this pattern is supposed to follow the history cursor pattern, where the sub panels attach to the closest history cursor panel..  hence the expedition/exploration etc attach a call back to the UCTG and the UCTG fires the call to them..

robbyxp1:
See reply in ticket, its working as it was designed at the moment.

robbyxp1:
Going to keep the pattern, thanks for the push, but this breaks it.

PR #3119, opened 2021-06-21
Tidying up

Eahlstan:
Doesn't actually change anything but makes adding and maintaining things easier.
(Well it now shows the planet type on every planet, so that could be called a change.)

Now I changed some wording:

Previously if js.Atmosphere was taken from "Atmosphere" it would say
"Has an thin neon rich atmosphere."
If "Atmosphere" was empty and js.Atmosphere was taken from "AtmosphereType" it would say
"Has an Ammonia Oxygen."

Now says
"Atmosphere: thin neon rich atmosphere."
and
"Atmosphere: Ammonia Oxygen."

Also: Volcanism and atmopshere are no longer shown on non-landble bodies as it just blowing the output up without any use.

`information.Append((js.IsStar) ? Bodies.StarName(js.StarTypeID) + "." : null);` fixes a really old bug: If you selected 'show all stars' but not 'show more information' only the body designation was shown without any hint what that body actually is.


robbyxp1:
So, your repeating translations and text which can be obtained from Bodies.cs - see PlanetTypeName.  I suggest you use them instead of repeating the translations

robbyxp1:
In fact, see if there are more repeats in the current code as well.

Eahlstan:
> So, your repeating translations and text which can be obtained from Bodies.cs - see PlanetTypeName. I suggest you use them instead of repeating the translations

Thanks for the hint.

PR #3134, opened 2021-07-12
Content adjustments to the Estimated Values panel

fredjk-gh:
1) Exclude 0-value things (like asteroid belts) from the table.
2) Strip the system name from the body names. This results in names that are more consistent with the Scan Panel (ie. using "Main Star" for the single primary star in the system) and shortens the value in the column leaving more room for other values and/or preventing text wrapping.

Combined, these minor changes make the table much easier to read at-a-glance. See attached examples.

EDDiscovery-EstValue-NameShortening_SingleStar

EDDiscovery-EstValue-NameShortening_MultipleStars


robbyxp1:
Going to have to remove the main star bit, its not translatable.. but keep the rest.

PR #3138, opened 2021-07-19
Move features to be shown on landables down two levels to avoid overloading the main level with too much options

Eahlstan:
Screenshot 2021-07-19 015908


robbyxp1:
Your conflating the pass criteria for showing something (the first two submenus) with what to show when a body is printed (the show ones). Would be better to group all the show entries under a Show.. submenu.

robbyxp1:
And too many submenus just makes it unwieldy btw, there is a balance to be had.

PR #3149, opened 2021-07-30
New event, new update trigger

Eahlstan:


robbyxp1:
Good point, but actually, we can do better than that.

robbyxp1:
See my method of doing it, on tip now. thanks

PR #3241, opened 2022-03-26
Include WebView2 loader; set user data folder

### LINKED ISSUE #3239: EDDiscovery 15.0.0.0 still using IE11 even after installing the WebView2 Runtime
Hello! After installing 15.0.0.0 and the WebView2 runtime through the link you provided in the release page, EDDiscovery still prompts that it is using IE11 whenever I use a tab that loads a web page (such as EDDB):
image

I also verified that the WebView2 Runtime is installed:
image

I'm running EDDiscovery on Win10 64-bit 21H1.

klightspeed:
This should fix #3239


robbyxp1:
I've come to the same conclusion. I'll cherry pick this.

PR #3258, opened 2022-04-29
Rename row

Eahlstan:
Organic scans row shows value of scans, so let's call it Values to be the same as other value rows


robbyxp1:
thanks

PR #3281, opened 2022-06-18
Add space and N0 to system value

Eahlstan:


robbyxp1:
will merge when all together.

PR #3286, opened 2022-07-10
Add biological count info to organics panel

Eahlstan:
Screenshot 2022-07-10 183519
Screenshot 2022-07-10 182954


robbyxp1:
Ok, there was a few errors, copied it in manually.

PR #3320, opened 2022-09-16
Ignore received texts in Surveyor

Eahlstan:
Stops the history cursor to mess up the system line


robbyxp1:
See latest purge, its bigger than just this Jim.

PR #3326, opened 2022-09-29
make datetime filter in organic scans panel respect display time setting

RadarCZ:


Eahlstan:
I see your point (I got an event happening shortly before midnight UTC that is on the next day in local time that doesn't show when selecting the local event day), but I think simply changing line 278 to

`orglist = orglist.Where(x => (start == null || x.Item2.EventTimeLocal >= start) && (end == null || x.Item2.EventTimeLocal <= end)).ToList();`

is:
- easier
- resembles the way other panels compare the event time
- does the same

I might be wrong and blind to see other effects your change has though.

RadarCZ:
My change was supposed to make even the inputs (on the organics panel) show DateTime in a format I choose in the settings. Gonna look into your suggested change though.

robbyxp1:
RadarCZ is right, because its better to store the value in the DB in UTC, and only display it in local..  thats so it works better over a daylight saving or change in TZ

Eahlstan:
Always nice to learn more 🙃

robbyxp1:
I'll pull this in later, thanks for the PR

RadarCZ:
And feel free to edit the version number, since the releases use standard SemVer, but C# has 4 numbers, which I probably also misused :D
Thanks for the notice

Eahlstan:
> And feel free to edit the version number, since the releases use standard SemVer, but C# has 4 numbers, which I probably also misused :D Thanks for the notice

It's more the pushed .gitmodules I'm worried to get pulled in, can you add that to your gitignore so you don't PR this in the future?

RadarCZ:
Now that i think about it, it doesn't really makes sense to be pushed, thanks for the "kick". Will fix right now!

robbyxp1:
I'll fix this in the same manner as all other panels.   So the current code is now wrong.

Pushing code down to the lower levels when they don't need to know about the upper level ideas of being able to handle date time pushed is not correct.  Handling the variations of date time is a EDD project level problem.

robbyxp1:
Thanks. Let me fix it, i have all the code to fix it in my mind.

RadarCZ:
Alright then, closing this.

Issue #3338, opened 2022-11-11
crash on opening 3d map ubuntu 22.04


adamthat:
### Exception Details:
>EDDiscovery v15.1.1.500 There was an unhandled UI exception.
>```
>==== BEGIN ====
>System.DllNotFoundException: opengl32.dll assembly:<unknown assembly> type:<unknown type> member:(null)
>  at (wrapper managed-to-native) GLOFC.GLStatics.wglGetCurrentContext()
>  at GLOFC.GLStatics.GetContext () [0x00000] in <8e95fc6b11c24186aef5eb01c2c9e02f>:0 
>  at GLOFC.WinForm.GLWinFormControl..ctor (System.Windows.Forms.Control attachcontrol, OpenTK.Graphics.GraphicsMode mode, System.Int32 major, System.Int32 minor, OpenTK.Graphics.GraphicsContextFlags flags) [0x001e4] in <8e95fc6b11c24186aef5eb01c2c9e02f>:0 
>  at EDDiscovery.UserControls.UserControl3DMap.Init () [0x0000d] in /home/adam/src/EDDiscovery/EDDiscovery/UserControls/3DMap/UserControl3DMap.cs:40 
>  at EDDiscovery.UserControls.UserControlCommonBase.Init (EDDiscovery.EDDiscoveryForm ed, System.Int32 dn) [0x0005b] in /home/adam/src/EDDiscovery/EDDiscovery/UserControls/UserControlCommonBase.cs:63 
>  at (wrapper remoting-invoke-with-check) EDDiscovery.UserControls.UserControlCommonBase.Init(EDDiscovery.EDDiscoveryForm,int)
>  at EDDiscovery.PopOutControl.PopOut (EDDiscovery.PanelInformation+PanelIDs selected) [0x000e3] in /home/adam/src/EDDiscovery/EDDiscovery/PopOuts.cs:120 
>  at EDDiscovery.EDDiscoveryForm+<>c__DisplayClass210_0.<buttonExtPopOut_Click>b__0 (System.Object s, System.EventArgs ea) [0x00001] in /home/adam/src/EDDiscovery/EDDiscovery/EDDiscoveryForm.cs:1433 
>  at ExtendedControls.ExtListBoxForm.listcontrol_SelectedIndexChanged (System.Object sender, System.EventArgs e) [0x0001f] in /home/adam/src/EDDiscovery/ExtendedControls/ExtendedControls/Forms/ListBoxForm.cs:110 
[...]

PR #3339, opened 2022-11-12
Disable 3D map under non-windows platforms

klightspeed:
This should work around the issue causing #3338


robbyxp1:
On hold - Bravada has a better solution on OFC which was be incorporated.

PR #3375, opened 2023-01-18
Galaxyselector translation text updated to match the actual text in t…

Eahlstan:
…he code.


Eahlstan:
There was a typo in the sorting drop down. Added fix to the PR.

PR #3383, opened 2023-02-01
Hide megaships older than 2 weeks to stop spam

Eahlstan:


EDDiscovery/UserControls/Overlays/UserControlSurveyor.cs
```diff
@@ -733,7 +733,7 @@ async private void CalculateThenDrawSystem(ISystem sys)
 
                     // mirrors scandisplaynodes
 
-                    var notexpired = signallist.Where(x => !x.TimeRemaining.HasValue || x.ExpiryUTC >= DateTime.UtcNow).ToList();
+                    var notexpired = signallist.Where(x => (!x.SignalName.Contains("-class") && (!x.TimeRemaining.HasValue || x.ExpiryUTC >= DateTime.UtcNow)) || (x.SignalName.Contains("-class") && x.RecordedUTC > DateTime.Now.AddDays(-14))).ToList();
```
robbyxp1, line 736: Should be DateTime.UtcNow.AddDays(-14)

PR #3402, opened 2023-04-03
Make ScanSummary hideable

Eahlstan:
I got many Surveyor popouts that are set to show very rare things, they weren't completely invisible and that annoyed me.

Screenshot 2023-04-03 230726
Screenshot 2023-04-03 230704
Screenshot 2023-04-03 230645


robbyxp1:
thanks

PR #3407, opened 2023-04-16
Fix import options and compass bookmarks.

Eahlstan:
CSV, Text File, JSON means csv, txt, json - not csv, csv, txt, json


Eahlstan:
Creating a surface bookmark with the compass didn't send that bookmark back to the compass. The full body name is to blame, stripping it off the system name makes it work as expected.

PR #3412, opened 2023-05-01
Fix misleading wording and Apex CAPI error

Eahlstan:
mk.IsValid returns true when it's valid, therefore
`MK is invalid IsTrue`
is misleading as it's not invalid.


robbyxp1:
Why dont we want to get station data when  in a taxi?

Eahlstan:
As I said in the commit: Taking a taxi from A to B and finally docking at B creates a docking event for B, but the CAPI still sends data from A. 
Never going to work, always will fill the log with error messages.

robbyxp1:
Lets talk over on discord

PR #3437, opened 2023-08-23
Making the Compass "Hide when no lat/long" option apply to the target lat/long inputs.

Identifier:
Currently, the compass doesn't auto-hide even when there's no target specified.  There's a checkbox in its settings, for "hide when no lat/long", and I assumed it meant "hide when there is no latitude or longitude target set" in the input boxes, but after looking at the source code I realize it means "hide when EDDiscovery can't detect a valid player position".  In any case, I'd really like to be able to hide the compass when I'm not using it (i.e. when I haven't input a lat/long), so I think this checkbox would be an intuitive way to do that.

### DIFF
--- a/EDDiscovery/Translations/EDDiscoveryTranslations.cs
+++ b/EDDiscovery/Translations/EDDiscoveryTranslations.cs
@@ -228,6 +228,7 @@ namespace EDDiscovery
         UserControlCompass_hidewheninSRV,
         UserControlCompass_hidewheninship,
         UserControlCompass_hidewhennolatlong,
+        UserControlCompass_hidewhennotarget,
 
 
         ImportExportForm, // Control 'Export'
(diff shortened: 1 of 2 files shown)


Eahlstan:
Greetings.

I do see why you want to hide it, but the way you implemented it will never get approved.

Your only change was on line 265 (thanks GitHub for marking every single line as deleted and readded -.- ):

````if (intransparent && IsSet(CtrlList.hidewithnolatlong) && !position.Location.ValidPosition)```

to

```if (intransparent && IsSet(CtrlList.hidewithnolatlong) && !(position.Location.ValidPosition && numberBoxTargetLatitude.IsValid && numberBoxTargetLongitude.IsValid))```

This surely works for your special case, but is a bad design for almost everyone else.
You'd need to add another line to the options drop down menu

Screenshot 2023-08-23 203134

and then change the current line 265 to hide it when one or more hiding options are selected.

Identifier:
Thanks for the feedback @Eahlstan 

I've added an another line in the settings like you suggested.  Can you take a look?

Eahlstan:
Looks better, but we'll have to wait for @robbyxp1 . He's the one who knows what he's doing, I was just giving hints on why he'd reject your first try :)

robbyxp1:
I'll have a look this weekend, thanks for the pr

PR #3438, opened 2023-08-24
Removed trailing space after pop-out window title.

Identifier:
When you create a pop-out window, the first one has a trailing space after its name.  This for some reason breaks tools like Desktop+ from re-detecting the window properly.  This pull-request simply removes the extra trailing space.


robbyxp1:
Looks good, thanks.

PR #3469, opened 2024-01-17
Date pickers allow for exact selection...

Eahlstan:
...so use the exact time selected and not start and end of the day.


Eahlstan:
Screenshot 2024-01-17 213108
Screenshot 2024-01-17 213049

PR #3480, opened 2024-02-08
Ungrid wording

Eahlstan:


robbyxp1:
View on History Panel is not right.. you could be in a splitter/grid with a "History grid view" and it won't swap to the history panel

            { new PanelInfo( PanelIDs.TravelGrid, typeof(UserControlTravelGrid), "History", "TravelHistory", "History grid view") },

https://github.com/EDDiscovery/EDDiscovery/wiki/Using-the-History-Grid-Panel

I'd use "View on nearest History Grid Panel"

PR #3493, opened 2024-03-09
Improve gravity display on Organics

Eahlstan:
Make gravity use the same precision as other panels, make it impossible to show 0 gravity on bodies with more than zero gravity by rounding away from zero, change G (gravitational constant) to g (gravitational acceleration)


Eahlstan:
Screenshot 2024-03-09 152217

PR #3502, opened 2024-04-05
Move galactic mapping and GEC requests off UI thread

klightspeed:
`HttpWebRequest.BeginRequestAsync` under Mono (both under Linux and under Wine / CrossOver) appears to use the message loop to perform HTTP request / response processing when executed on the UI thread.

Move the EDSM galacticmapping.json and EDAstro GEC gecmapping.json requests into a task to get them off the UI thread so they do not attempt to process the HTTP request / response using the message loop.

### DIFF
--- a/EDDiscovery/EDD/EDDiscoveryForm.cs
+++ b/EDDiscovery/EDD/EDDiscoveryForm.cs
@@ -28,6 +28,7 @@ using System.Drawing;
 using System.Globalization;
 using System.IO;
 using System.Linq;
+using System.Threading.Tasks;
 using System.Windows.Forms;
 
 namespace EDDiscovery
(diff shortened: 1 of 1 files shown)


robbyxp1:
This does not work, the files must be present before the code moves on, as the files are then used to construct GMO info this only occurs on first instance of the program as well. Thats why its a blocking action.

So rejected.

klightspeed:
I'm guessing you didn't see the .Wait()?

robbyxp1:
Why does this improve this with wait, vs the blocking form, on Linux? 

Also, this form of code is used in other places, is that affected by the linux implementation.

klightspeed:
Obsoleted by https://github.com/EDDiscovery/BaseUtilities/pull/38

PR #3572, opened 2024-10-07
Harmonise number formatting and add missing units to ScanGrid

Eahlstan:
From

Screenshot 2024-10-07 184436

to

Screenshot 2024-10-07 184021


robbyxp1:
Noted, will bring in when it all gets merged..

PR #3609, opened 2024-12-19
[85865651] allow systems on bookmarks to be sent to expedition panel

FlynnDynamics:


FlynnDynamics:
https://github.com/user-attachments/assets/507040ba-5095-42b3-af2d-f2b3932993b3

robbyxp1:
The Add to stuff looks good.

The enumcms is not needed. This user control is a special case, where I cheated and reused the translations IDs of another panel and manually translated it. You can see that in line 63-66.  You just need to add one more ID for the new right click entry and add a line after 66 to translated that control text.

And the translation-example-uc.ylp needs updating with the new ID.

FlynnDynamics:
Should fit now.

PR #3631, opened 2025-02-02
Update french.

Etienne1990:
I used ingame words at the maximum when i could find them.
There is still some missing translations.


robbyxp1:
Thank you - i'll bring this in later in the release cycle!

robbyxp1:
Into master, not 18.1.7

PR #3711, opened 2025-08-19
Improve Chinese translation

LuckyZone:


robbyxp1:
I'm not ignoring this btw, its just not the right time yet to bring it in!

LuckyZone:
As long as the data format is correct, I will continue to improve the translation content. However, without convenient local debugging tools, it is difficult to guarantee the quality of the translation.

PR #3758, opened 2025-11-14
Build debug and release in appveyor

klightspeed:
Test both debug and release builds in Appveyor.
Built portable zip is still release.


robbyxp1:
Thanks

PR #3795, opened 2026-01-04
Pp-merits-cycle-date

shufton:
Fix incorrect start date for cycles
Update Cycle start calculation to be from 07:00 UTC to allow correct allocation of merits 00:00 -> 06:59


robbyxp1:
I've fixed up your previous push onto the 19.x branch, and i'm now going to move those files directly first to master first so I know they are synced.  I need to do this first before I can accept any PRs on this.  So bear with and update once i've done that.

Closing as not possible to bring in without confusing me.

robbyxp1:
Ok master has been updated from 19, and i've applied your cycle start date fixed.  I've also enabled translation on this panel.

Sorry I could not just pull it in by I need to know the state of play between 19 and master at the moment.

thanks

shufton:
@robbyxp1 Hi, I appreciate you sorting this out for me, apologies that I made a mess - I didn't realise gh made a mess of that merge commit. Will take more care in future.

PR #3821, opened 2026-03-22
Detect FDev Duplicate PowerplayMerit events

shufton:
FDev in their infinite wisomd can occaisionally send duplicate PowerplayMerit events, in one of two types.

The first type is a repeat identical message (type 1) and the second one where the duplicate comes in with an updated total (type 2),. A subsequent message is then received,  with the merit total increased with the subsequent award amount, minus the previous duplicate value. I have been able to account for all of these, except in one scenario, where the type 2 duplicate is the last merit grant in the stream. For this, the only way to confirm is on the next startup with the "powerplay" event. In this case I have chosen to display a warning about this in the page, warning the CMDR that there may be a duplicate merit that can only be checked via a re-log.

### DIFF
--- a/EDDiscovery/UserControls/CurrentState/UserControlPPMerits.cs
+++ b/EDDiscovery/UserControls/CurrentState/UserControlPPMerits.cs
@@ -554,21 +554,10 @@ namespace EDDiscovery.UserControls
                 if (pendingMeritsQueue.Count == 0)
                     break;
 
-                if (!allowTailCommit && pendingMeritsQueue.Count == 1)
-                {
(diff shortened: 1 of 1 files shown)


shufton:
Hey, I've been going through this in my head as I was sure there was an edge case I missed - I think I have but still not clear so going to try and reproduce. I can't decide if there is a situation where the type 1 (duplicate award and total  message) can be included if it gets interleaved with a type 2 one.  I also want to try and break the algorithm and see how it recovers, though no log I've replayed has so far. Hold off on merging this for now - if you want to close to keep things clean I'll resubmit another PR with any updates.

robbyxp1:
Okay, i was going to merge in in this week, i'll wait for you.

Issue #3856, opened 2026-07-09
Suggestion - Star Type in Nearest Star List


sjgarth:
Nearest star list can be useful, but in my mind it would be more useful if we were able to identify it as a fuel star quickly and easily.
Consider adding a star type column in the nearest star table.

robbyxp1:
Once the code is all together again i'll pull the PR in, thanks

PR #3859, opened 2026-07-11
Add StarType to nearest stars panel

klightspeed:
Requested in #3856


robbyxp1:
Thanks, i'll bring it in on next next mega merge

